PR #86, opened 2026-05-10
fix(asi-import): hotfix for ASI plugin installation issue in variant-pack mods

Democles85:
This release addresses a critical issue where companion .asi plugins bundled with variant-pack mods were not being installed correctly. The cleanup of the pre-extract directory is now deferred until after the ASI install loop, ensuring that the necessary files are available for copying.


faisalkindi:
Thanks for catching this and writing the patch. I re-applied the same fix on top of v3.3.0 master in db1eabf so the version bump in your branch (3.2.17) does not conflict with the v3.3.0 release that already shipped. The cleanup now defers the pre-extract rmtree until after the ASI install loop runs, exactly as you proposed. Closing this in favor of the direct commit, but the credit lands in the next release notes.

PR #100, opened 2026-05-11
feat(mods-page): bulk Uninstall Selected button + Esc clears selection + 16-locale i18n

Democles85:
## Summary

Surfaces the existing batch-uninstall flow as a first-class
**Uninstall Selected** button in the Mods page header (next to **New
Folder** and **View Conflicts**), adds an **Esc** keyboard shortcut
that clears visual selection on the Mods page, and ships translations
for the three new strings across all 16 supported locales.

## Motivation

The right-click menu has always carried an `Uninstall N mods` batch
action that operates on Ctrl+Clicked cards, but the existence of that
flow was discoverable only by accident — there was no visible UI
affordance saying "yes, you can bulk uninstall". Users wanting to
clean up and start fresh kept asking for a "delete all mods" button.

The new header button reuses the existing `_ctx_batch_uninstall` path,
so the confirmation dialog, the disable → Apply (revert game files) →
remove-from-DB pipeline, and the already-disabled fast path are shared
with the right-click menu — there is no parallel destructive code path.

## What's in this PR

- **`src/cdumm/gui/pages/mods_page.py`**
  - `_uninstall_selected_btn` PushButton placed between New Folder and
    View Conflicts in the header row, with red styling matching View
    Conflicts and a disabled-state palette.
  - `_get_uninstall_target_ids()` returns the **union** of
    (per-card enable-checkbox is ticked) and (card is visually
    highlighted via Ctrl/Shift/plain click). An earlier draft used a
    priority rule (visual selection wins when present) which created
    a sticky-mode bug — plain-clicking a card body silently captured
    the button onto the visual set and ignored later checkbox toggles
    until the user double-tapped Select all to clear the highlight.
    Union dodges this entirely.
  - `_sync_uninstall_selected_btn()` keeps the live label
    (`Uninstall N selected`) and enabled state in sync, hooked into
    every selection-changing path (`_on_select_all`, `_on_card_clicked`,
    `_on_ctrl_a`, `_deselect_all_cards`, `_on_mod_toggled`,
    `_on_select_all_in_group`, `_on_context_menu_requested` pre-select
    branch, and `refresh`).
  - `_on_uninstall_selected()` delegates straight to
    `_ctx_batch_uninstall(target_ids)` so the dialog/flow are shared.
  - **Esc** shortcut wired to `_deselect_all_cards()` with
    `Qt.ShortcutContext.WidgetWithChildrenShortcut` so it does not
    compete with the search box's clear-on-Esc, inline rename's
    commit-on-Esc, or any other Esc handler in the window. Clears the
    transient visual highlight only — the per-card enable-checkbox is
    intentionally untouched (persistent game state vs. selection
    gesture).
  - `retranslate_ui()` re-runs the sync helper and resets the tooltip
    so the dynamic count label and tooltip both retranslate on locale
    switch.

- **`src/cdumm/translations/*.json` (16 files)**
  - Three new keys added immediately after `mod_list.select_all`,
    preserving the existing 2-space indentation:
    - `mod_list.uninstall_selected` — base label.
    - `mod_list.uninstall_selected_n` — count-aware label with
      `{count}` placeholder.
    - `mod_list.uninstall_selected_tooltip` — explains the dual
      ticked-or-highlighted selection model.
  - Each language's verb matches the existing uninstall vocabulary
    already used elsewhere in the same file (its
    `context_menu.uninstall_multi` string), so the button reads
    consistently with the existing right-click menu translation in
    that locale.
  - Plural-genitive forms in Polish and Russian follow the shape their
    existing `context_menu.uninstall_multi` strings already use.
  - Native-speaker review of the tooltip wording is welcome.

- **`src/cdumm/__init__.py`** — `__version__` bumped 3.3.2 → 3.4.0
  (semver minor: new user-facing feature, no API breakage, no schema
  migration).

- **`src/cdumm/gui/changelog.py`** — `v3.4.0` entry at the top of
  `CHANGELOG`, with the same narrative voice used in existing
  entries, covering the button, the union selection model, the Esc
  shortcut, and the 16-locale translation rollout.

## Why a union (not a priority)

The very first draft of the button preferred visual selection when
present, on the reasoning that power users with Ctrl+Click muscle
memory deserved exact control over what the button targets. In
testing this created a clear sticky-mode bug:

1. User plain-clicks a card body → that card becomes the single
   visually-selected card, and the priority rule pins the button onto
   just that card.
2. User then ticks another card's checkbox to add it to the batch →
   nothing happens to the button, because visual selection still
   wins.
3. The only escape is to double-tap **Select all** to clear the
   visual highlight, which is hidden, opaque behaviour.

The shipped logic treats both inputs as a union — the count is always
explainable as "ticked OR highlighted", every checkbox edit always
updates the button, and Ctrl+Click on top of an existing checkbox set
just adds more. The `_get_uninstall_target_ids()` docstring records
the bug-and-fix reasoning so the design choice survives future
refactors.

## Esc semantics

Esc clears **only** the transient visual highlight (the
`set_selected(False)` path on each card). The per-card enable-
checkbox state is intentionally untouched. The reasoning:

- Visual highlight is a **gesture** ("I'm picking these for the next
  batch action") — clearing it is harmless and matches every other
  multi-select UI.
- Enable-checkbox is **persistent state** — flipping it changes
  pending Apply state and can revert or install game files. Esc is
  an "abort selection" key, not a "revert state" key, so it must not
  touch this.

The shortcut uses `WidgetWithChildrenShortcut` to keep it scoped to
the Mods page's widget tree, avoiding collisions with `SearchLineEdit`
clear-on-Esc, `QLineEdit` inline-rename commit-on-Esc, or any other
window-level Esc handler.

## UX discussion captured

A short UX side-question came up during development and is worth
flagging here for upstream context:

> Should Ctrl/Shift+Click also tick the enable-checkbox?

Decision: **no**, kept separate. The two have semantically different
meanings (enable-checkbox = persistent game state vs. highlight =
transient batch gesture). Conflating them at the per-click level
would cause obvious accidents — e.g. a Shift+Click range selecting 15
cards would also flip 15 enable-checkboxes, half of which would land
on the wrong state and silently queue an Apply-to-Deactivate. The
existing `_on_select_all` already conflates the two for the master
checkbox, which is an unambiguous "I want everything" command; that
same conflation is the wrong default for per-card clicks. The button's
union logic already lets both inputs feed it without needing to flip
checkbox state behind the user's back.

## Test plan

- [x] Open the Mods page on a real install with a non-zero mod count.
- [ ] Confirm the new **Uninstall selected** button sits between
      **New Folder** and **View Conflicts**, greyed out at rest.
- [ ] Tick a single checkbox → label updates to
      `Uninstall 1 selected`, button becomes clickable. Untick → back
      to disabled.
- [ ] Tick several → live count tracks ticks/unticks 1:1.
- [ ] Ctrl+Click cards on top of a partial tick set → count grows by
      one per click; Ctrl+Click an already-highlighted card → it
      drops out of the highlight portion (but stays if it was also
      ticked).
- [ ] Plain-click a card body → visual selection switches to that
      card; count = (still ticked set) ∪ {that one card}. Subsequent
      checkbox toggles continue to update the count (regression test
      for the union-not-priority fix).
- [ ] Click **Select all** → all checkboxes tick + all cards
      highlight; count = total visible mod count.
- [ ] Press **Esc** → all visual highlights clear; checkboxes
      unaffected; count drops to "still ticked" total.
- [ ] Click the button on a non-empty selection → standard
      confirmation dialog (`Remove N selected mods? This cannot be
      undone.`); **Cancel** = no-op, **OK** = standard disable →
      Apply → remove pipeline; already-disabled mods skip the Apply.
- [ ] Switch UI language (Settings → Personalization → Language)
      across at least two locales; confirm button label, count form
      (`Uninstall N selected` localised), and tooltip retranslate
      immediately.

## Notes

- The button's red palette deliberately matches **View Conflicts** so
  the two destructive/warning-coloured affordances visually cluster.
- All sync hooks are idempotent and cheap (single list comprehension
  over `_mod_cards`), so the extra calls on selection-changing paths
  have negligible overhead.
- I am happy to drop the `release(v3.4.0)` commit and the
  `__version__` / `CHANGELOG` edits if upstream prefers to manage
  versioning independently — let me know and I will rebase.


Democles85:
Actually I think I need to change the select checkbox behaviour. It shouldn't trigger Uninstall All button. Selecting the cards should.

faisalkindi:
Hi, working through the open PRs now. Three things would unblock a merge from my side.

1. This branch is well past current master at this point (we are staging v3.3.15 from v3.3.2 when you opened). Could you rebase against latest master?

2. Two commits carry Co-authored-by: Cursor <cursoragent@cursor.com>. The project keeps a strict no-AI-attribution policy on contributions, so those trailers need to come off. Interactive rebase with message edits, or a single squashed commit without the Cursor trailer, both work.

3. The release(v3.4.0) commit bumps __version__ from 3.3.2 to 3.4.0. Versions are cut manually at tag time, so please drop that commit and let me bump the version when the merged change actually ships.

Happy to handle the rebase myself if you would rather I cherry-pick the substantive commits and drop the rest, just say the word.

PR #122, opened 2026-05-14
feat(skill): add Mission Efficiency primitive field schema entries

voiddoivvoid:
## Summary

Adds a complete set of `field_schema/skill.json` primitive entries for Mission Efficiency skill values.

This expands the existing `skill.pabgb` Format 3 primitive support from the initial Farming I example into the full Mission Efficiency target set, allowing Mission Efficiency `.field.json` mods to use typed primitive intents instead of `_buff_data_raw` blob replacement.

## What this changes

Adds Mission Efficiency primitive schema entries for:

- Farming I / II / III
- Ranching I / II / III
- Logging I / II / III
- Mining I / II / III
- Forging I / II / III
- Strengthening I / II / III
- Fishing I / II / III
- Gathering I / II / III
- Banking I / II / III
- SuperWorker

Each schema entry maps a readable Format 3 field name such as `mission_eff_farming_i` to the correct tagged primitive location inside `skill.pabgb`.

Uses `tid` + `value_offset: 5` with `type: "i64"` for each Mission Efficiency scalar value.

## Why

Mission Efficiency values are stored as tagged integer primitives inside `skill.pabgb`.

Before this, Mission Efficiency `.field.json` support required `_buff_data_raw` replacement, which works but is more brittle because it depends on larger raw byte blobs.

These schema entries allow the mod to target the actual primitive values directly through CDUMM's Format 3 field-schema path.

This makes the generated `.field.json` format cleaner, smaller, easier to review, and easier to regenerate after game updates.

## Validation

Generated these entries from the current `skill.pabgb` / `skill.pabgh` layout and tested the resulting CDUMM build locally.

The generated primitive `.field.json` version of Mission Efficiency imports and applies as intended in the local build.

## Notes

This PR only updates `field_schema/skill.json`.

It does not change the Format 3 handler or apply pipeline logic. The existing handler already supports resolving skill primitives through `field_schema/skill.json`; this PR simply adds the missing Mission Efficiency schema entries.

### DIFF
--- a/field_schema/skill.json
+++ b/field_schema/skill.json
@@ -1,10 +1,51 @@
 {
-  "_credit": "Initial entries for skill.pabgb primitive Format 3 mods contributed by voiddoiv via Nexus comments on 2026-05-08. CDUMM ships these as community-curated values; correctness verified against author's working build, not independently re-derived from a vanilla skill.pabgb dump.",
-  "_format": "JMM-compatible field_schema. Each entry maps a Format 3 'field' name to a tid (0xNNNNNNNN little-endian type tag the engine writes before the value), value_offset (bytes from the tid to the actual value), and type (i8/u8/i16/u16/i32/u32/f32/i64/u64/f64).",
-  "_contribute": "Open a PR at https://github.com/faisalkindi/CrimsonDesert-UltimateModsManager with new entries for skill primitives you have verified.",
-
-  "mission_eff_farming_i": {
+  "worker_skill_farming": {
     "tid": "0x0F427600",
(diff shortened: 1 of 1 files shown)


voiddoivvoid:
Paused to reduce schema redundancy and revise field names so they describe the underlying skill primitive more accurately.

voiddoivvoid:
## Update

Reworked the schema before review:

- Reduced redundant per-tier entries into grouped `worker_skill_*` fields.
- Renamed fields to describe the underlying worker skill primitive instead of the Mission Efficiency mod.

PR #124, opened 2026-05-14
feat(skill): add skill primitive schema and indexed TID selection

voiddoivvoid:
## Summary

Adds `skill.pabgb` primitive schema entries and supports optional `tid_index` selection for repeated TID matches.

## Changes

- Add worker skill primitive fields for worker skill style mods.
- Add stamina and spirit primitive fields for stamina and spirit style mods.
- Add optional `tid_index` parsing on Format 3 intents.
- Add optional `tid_index` support in `field_schema` entries.
- Pass `tid_index` through the Format 3 apply path.
- Preserve the existing safe behavior when `tid_index` is not supplied.

## Why

Some skill primitives are uniquely resolved by `entry`, `key`, and `tid`.

Other skill primitives can use the same TID more than once inside one entry. In that case, CDUMM should not guess which match to patch. `tid_index` provides an explicit zero-based selector for those repeated primitive matches.

## Validation

- Existing single-match TID behavior remains unchanged.
- Multi-match TID behavior still refuses to guess when `tid_index` is absent.
- `tid_index` selects the intended repeated TID occurrence when supplied.

## Known Limitation

`tid_index` is less durable than a fully unique TID because it depends on repeated TID order remaining stable.

It is included for cases where the best available primitive marker is reused inside the same entry.

## Merge note

The worker skill fields and cost fields are safe without `tid_index` support:

- `worker_skill_*`
- `skill_stamina_cost`
- `skill_spirit_cost`

The delta fields should only be merged if optional `tid_index` support is also merged:

- `skill_stamina_delta`
- `skill_spirit_delta`

Those two TIDs can appear multiple times inside the same `skill.pabgb` entry. Without `tid_index`, CDUMM correctly refuses to guess which repeated TID occurrence should be patched.

### DIFF
--- a/field_schema/skill.json
+++ b/field_schema/skill.json
@@ -1,4 +1,8 @@
 {
+  "_credit": "Initial entries for skill.pabgb primitive Format 3 mods contributed by voiddoiv via Nexus comments on 2026-05-08. CDUMM ships these as community-curated values; correctness verified against author's working build, not independently re-derived from a vanilla skill.pabgb dump.",
+  "_format": "JMM-compatible field_schema. Each entry maps a Format 3 'field' name to a tid (0xNNNNNNNN little-endian type tag the engine writes before the value), value_offset (bytes from the tid to the actual value), and type (i8/u8/i16/u16/i32/u32/f32/i64/u64/f64).",
+  "_contribute": "Open a PR at https://github.com/faisalkindi/CrimsonDesert-UltimateModsManager with new entries for skill primitives you have verified.",
+
   "worker_skill_farming": {
     "tid": "0x0F427600",
     "value_offset": 5,
--- a/field_schema/skill.json
+++ b/field_schema/skill.json
@@ -48,5 +48,25 @@
     "tid": "0x0F427500",
     "value_offset": 5,
     "type": "i64"
+  },
+  "skill_stamina_cost": {
+    "tid": "0x0F425A03",
+    "value_offset": 5,
+    "type": "i64"
+  },
+  "skill_spirit_cost": {
+    "tid": "0x0F425B03",
+    "value_offset": 5,
+    "type": "i64"
+  },
+  "skill_stamina_amount": {
+    "tid": "0x0F425A00",
+    "value_offset": 5,
+    "type": "i64"
+  },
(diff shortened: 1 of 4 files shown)


faisalkindi:
Hi, working through the open PRs now. The branch is currently CONFLICTING against master. Could you rebase against latest master?

The scope makes sense. Your own merge note flagged that worker_skill_*, skill_stamina_cost and skill_spirit_cost are safe without tid_index, and the delta fields should ship paired with tid_index support. If you would rather scope-reduce to just the safe fields and land the delta+tid_index work in a follow-up, that also works.

Happy to rebase or scope-reduce myself if you would prefer, just say the word.

voiddoivvoid:
Hey, thanks for checking it.

I’m not actively maintaining this branch/PR at this time, so feel free to rebase it, scope-reduce it, or restructure it however you think is best.

I did take another look, and the conflict looks localized rather than major. 

For the TID side: I don’t know the current state of the TIDs anymore, but in my testing they survived 3 game updates, so they were fairly stable at the time.

One disclaimer: the safe subset is probably the worker skill fields plus `skill_stamina_cost` / `skill_spirit_cost`. But stamina-related mods may be incomplete if only the cost fields land and the delta fields are left out. The delta fields need the `tid_index` support, or someone would need to maintain a v2 JSON path for those fields properly.

So feel free to rebase, reduce scope to the safe fields, or land delta + `tid_index` as a follow-up if that is cleaner.

faisalkindi:
Really appreciate you putting this together and being honest about the bandwidth. The TID stability data you shared is genuinely useful, three game updates without breakage is a stronger track record than I was assuming when I first looked at the indexed selector idea.

Since the branch is dirty against current master and you are stepping back from active maintenance, I am going to close this PR rather than leave it open indefinitely. The safe subset you outlined, worker skill fields plus skill_stamina_cost and skill_spirit_cost, is a reasonable scope and I will pick that up as a fresh maintainer commit when I get back to the skill schema work. Your name will be on it as the original concept author. The delta fields and tid_index path are deferred for now, those still need design work I am not ready to commit to.

Closing for that reason, not because the work was wrong. Thanks again for the contribution.

PR #226, opened 2026-06-28
fix: add missing en/de translation keys for Update All + ReShade Remove buttons (#223)

AgentKush:
Fixes #223.

## Problem

`test_all_literal_tr_keys_resolve_in_en` (in `tests/test_i18n_key_parity.py`) fails on v3.4.1:

```
tr() keys with no en.json entry (would render as raw keys):
  src/cdumm/gui/pages/mods_page.py: mod_list.update_all
  src/cdumm/gui/pages/mods_page.py: mod_list.update_all_running
  src/cdumm/gui/pages/reshade_page.py: reshade.remove_btn
```

v3.4.1 added the "Update All" button (`mod_list.update_all` / `mod_list.update_all_running`) and the ReShade "Remove" button (`reshade.remove_btn`), but their entries were never added to the translation files. Besides the failing build check, German users get the English fallback labels because the code falls back to the literal when the key is missing.

## Fix

Add the three keys to both `en.json` and `de.json`, placed next to their sibling keys. The added German values are pure ASCII, so no existing umlauts are touched (verified: non-ASCII byte count of `de.json` unchanged at 1730).

## Verification

Reproduced on a clean clone of `master`, applied the fix, ran the whole file:

```
tests/test_i18n_key_parity.py ......  [100%]
6 passed
```

All three invariants hold: every audited `tr()` key resolves in en.json, en/de key sets are identical (1300 keys each), and placeholder parity holds (these values have no placeholders).

PR #228, opened 2026-06-28
fix: get Windows CI green again (parallelize tests + raise cap + add missing i18n keys)

### LINKED ISSUE #223: `test_all_literal_tr_keys_resolve_in_en` failure
With the recent v3.4.1 update I am starting to get a new build test failure that was passing last time:
```py
=================================== FAILURES ===================================
____________________ test_all_literal_tr_keys_resolve_in_en ____________________

en = {'_language_name': 'English', 'about.app_title': 'Crimson Desert Mod Manager', 'about.bug_desc': 'Found a problem? Open an issue on GitHub.', 'about.bug_title': 'Report a Bug', ...}

    def test_all_literal_tr_keys_resolve_in_en(en):
        missing: list[str] = []
        for rel in AUDITED_FILES:
            text = (REPO / rel).read_text(encoding="utf-8")
            keys = {m.group(2) for m in _TR_RE.finditer(text)}
            if rel.endswith("conflict_view.py"):
                keys |= {m.group(1) for m in _LEVEL_KEY_RE.finditer(text)}
            for key in sorted(keys):
                if key not in en:
                    missing.append(f"{rel}: {key}")
>       assert not missing, (
            "tr() keys with no en.json entry (would render as raw keys):\n"
            + "\n".join(missing))
E       AssertionError: tr() keys with no en.json entry (would render as raw keys):
E         src/cdumm/gui/pages/mods_page.py: mod_list.update_all
E         src/cdumm/gui/pages/mods_page.py: mod_list.update_all_running
E         src/cdumm/gui/pages/reshade_page.py: reshade.remove_btn
E       assert not ['src/cdumm/gui/pages/mods_page.py: mod_list.update_all', 'src/cdumm/gui/pages/mods_page.py: mod_list.update_all_running', 'src/cdumm/gui/pages/reshade_page.py: reshade.remove_btn']

tests/test_i18n_key_parity.py:81: AssertionError
=============================== warnings summary ===============================
tests/test_find_7z_registry.py:11
  /build/source/tests/test_find_7z_registry.py:11: SyntaxWarning: invalid escape sequence '\S'
    - HKEY_LOCAL_MACHINE\SOFTWARE\7-Zip\Path  (admin install)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
FAILED tests/test_i18n_key_parity.py::test_all_literal_tr_keys_resolve_in_en - AssertionError: tr() keys with no en.json entry (would render as raw keys):
==== 1 failed, 1855 passed, 102 skipped, 1 deselected, 1 warning in 30.20s =====
```

AgentKush:
## What was broken

The Windows CI job was failing **two** ways at once (the second was hidden by the first):

1. **Timeout.** The full suite (1,721 tests) runs single-threaded in ~25-30 min on the 2-core runner and was cancelled at the 15-min cap. The maintainer's "<5s" figure is a local macOS dev box; the GitHub runner is far slower for the Qt GUI tests.
2. **A real test failure underneath it** — `test_i18n_key_parity.py::test_all_literal_tr_keys_resolve_in_en` fails because v3.4.1 added three `tr()` keys (`mod_list.update_all`, `mod_list.update_all_running`, `reshade.remove_btn`) without `en.json`/`de.json` entries. CI never surfaced it because the timeout killed the run first. (Same fix as #223 / #226.)

## What this does

- **Parallelize the suite** with `pytest-xdist` (`-n auto --dist worksteal`). `worksteal` lets idle workers steal queued tests so no single worker bottlenecks. This drops the wall-clock from ~27 min to **~14 min** (verified on the runner).
- **Raise `timeout-minutes` 15 -> 20** for a safe margin over that ~14 min. (Parallelization alone lands right at the old 15-min cap, so the bump is needed for reliable green.)
- **Add the three missing translation keys** to `en.json` and `de.json` so the i18n parity test passes.

## Verification

Run on the fork's Windows runner: the full suite **completes** — 1,664 passed, 57 skipped — in ~14 min, green, with comfortable headroom under the 20-min cap. (Before: cancelled at 15 min having only reached ~half the suite.)

## Honest note on speed

This makes CI **green and reliable**, not yet *fast*. ~14 min is dominated by a tail of slow Qt-render tests (`config_panel`/`preset_picker`/`reshade` dialog tests) that stand up real widgets and pump the event loop — that's inherently slow on the headless 2-core runner and is the critical path. Genuinely getting it to a few minutes would mean profiling those specific tests (`--durations`) and either speeding them up or splitting them, or moving to a 4-core larger runner. Happy to do that as a follow-up.

Closes #223.


AgentKush:
Closing for now — raising the cap doesn't actually make the suite fast, which is the real goal. Reworking this to fix the slow tests themselves; will reopen a proper PR once the suite is genuinely fast (it'll still carry the #223 key fix and the parallelization).

PR #239, opened 2026-07-01
tools: add cd-game-index (searchable catalog of a CD install)

AgentKush:
﻿## What

`tools/cd-game-index/` — a small standalone tool that turns a **Crimson Desert install** into a **searchable SQLite catalog** of everything the game ships. It does **not** change CDUMM itself; it reuses CDUMM''s `paz_parse` so the index matches what the manager sees.

On a current install: **1,660,550 assets across 33 archives + 266 keyed game-data tables, built in ~15s** into a ~400 MB SQLite. Read-only; it stores metadata (paths, IDs, sizes, locations), never asset bytes.

## How it works

CD ships content in numbered archive folders (`NNNN/0.pamt` index + `N.paz` blobs); every file is a PAMT entry recording its path, which `.paz` holds it, offset, sizes, and flags. The tool:

1. Finds every `NNNN/` with a `0.pamt`.
2. Parses each with `cdumm.archive.paz_parse.parse_pamt` (the same parser CDUMM uses).
3. Writes one row per entry into an `assets` table, and every `.pabgb`/`.pabgh` into a `data_tables` table.
4. Adds indexes so lookups by path / extension / category / archive are instant.

`--items` additionally tries to decode `iteminfo` records via the vendored `crimson_rs` (best-effort — documented as version-dependent; the catalog doesn''t depend on it).

## How to use

```sh
# from inside the repo (paz_parse auto-located):
python tools/cd-game-index/cd_data_index.py "E:/SteamLibrary/steamapps/common/Crimson Desert" --out cd_gamedata.sqlite
```

Example queries:

```sql
-- where does a file live?
SELECT path, archive, paz_file, offset FROM assets WHERE path LIKE '%iteminfo%';
-- every loading-scene sequencer file:
SELECT path, archive, orig_size FROM assets WHERE ext = '.paseq';
-- biggest game-data tables (item/NPC/quest/skill DBs):
SELECT name, orig_size FROM data_tables ORDER BY orig_size DESC;
```

Full schema, more examples, and the item-record limitation are in `tools/cd-game-index/README.md`.

## Why

There''s no public index of CD''s game files; this gives modders a fast "what''s in the game / where does asset X live" lookup, and maps all 266 keyed tables as the roadmap toward an item/skill/quest ID database. Optional tool under `tools/` — nothing in the app depends on it.

## Testing

Run end-to-end against a real 1.12.02 install using the in-repo default path: 1,660,550 assets / 266 tables / ~15s / 414 MB; verified the DB answers real lookups (search by name, by extension, largest tables, per-archive category counts).

### DIFF
--- a/tools/cd-game-index/README.md
+++ b/tools/cd-game-index/README.md
@@ -45,17 +45,22 @@ through the vendored `crimson_rs` parser. This is best-effort — see
 
 ## Usage
 
+The install folder is **auto-detected** (Steam / Epic / Xbox / macOS / Linux) via
+CDUMM's own game-finder, so you don't need to know your path — but you can pass
+one to override (e.g. a second copy, or if detection misses it).
+
 ```sh
-# From inside the repo (paz_parse auto-located):
-python tools/cd-game-index/cd_data_index.py \
-    "E:/SteamLibrary/steamapps/common/Crimson Desert" \
-    --out cd_gamedata.sqlite
+# From inside the repo — auto-detects your install:
+python tools/cd-game-index/cd_data_index.py --out cd_gamedata.sqlite
+
+# Point it at a specific install folder instead:
+python tools/cd-game-index/cd_data_index.py "D:/SteamLibrary/steamapps/common/Crimson Desert" --out cd_gamedata.sqlite
 
 # Also attempt item-record extraction:
-python tools/cd-game-index/cd_data_index.py "<game dir>" --out cd_gamedata.sqlite --items
+python tools/cd-game-index/cd_data_index.py --out cd_gamedata.sqlite --items
 
-# Running a standalone copy of the script:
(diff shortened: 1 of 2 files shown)


AgentKush:
Updated: the install folder is now **auto-detected** via CDUMM's `find_game_directories()` (Steam/Epic/Xbox/macOS/Linux), so the path argument is optional — it no longer assumes a specific drive/layout. Verified running with no path: auto-detected the Steam install and indexed 1,660,550 assets / 266 tables. You can still pass a path to override.

AgentKush:
Closing in favour of #242. The in-app **Game Data tab** (#242) is the intended deliverable "nothing separate" - this standalone CLI catalog lives outside the app and isn't needed for the in-app browser. Closing to keep one self-contained PR.

Issue #247, opened 2026-07-04
[Bug]: Error activating mods (CD 1.13.00 - CDUMM 3.4.2)


falobos76:
### Type of problem

A mod won't install / only partly applies

### Game install source

Steam

### CDUMM bug report

```text
--- SEVERITY: Bug (wrong behavior) ---

============================================================
CRIMSON DESERT ULTIMATE MODS MANAGER -- BUG REPORT
============================================================
Generated: 2026-07-05 00:01:48
App Version: 3.4.2

--- TL;DR ---
  2 issue(s) detected:
    1. Game exe was modified (2026-07-04 22:38) after your snapshot (2026-07-04 21:23) — game likely updated or verified. Revert mods and Rescan Game Files.
    2. Top overlap: 11 enabled mods all touch '0012/4.paz'. Whoever has the highest priority wins — re-order if the wrong one is winning.
  OK:
    - Not running as admin (good)

--- SYSTEM ---
OS: Windows-11-10.0.26200-SP0
Python: 3.13.12 (frozen)
Running as administrator: no
Game Dir: D:\GIOCHI\Steam\steamapps\common\Crimson Desert (exists)
Game Platform: Steam
Game Exe: D:\GIOCHI\Steam\steamapps\common\Crimson Desert\bin64\CrimsonDesert.exe (345.7 MB, mtime 2026-07-04 22:38)
Game Exe Compat Flags: none
Game Version Hash: c5573f24bf24e826
Game Drive Free: 282.4 GB free of 931.5 GB
App Data: C:\Users\Briz\AppData\Local\cdumm
App Data Size: 16.0 MB
App Data Drive Free: 685.0 GB free of 930.6 GB

--- STORAGE ---
CDMods: D:\GIOCHI\Steam\steamapps\common\Crimson Desert\CDMods
  vanilla/: 63 files, 4.47 GB
  deltas/: 342 files, 124.5 MB
  sources/: 374 files, 22.35 GB
  overlay/: (not present)

--- SETTINGS ---
  Theme: dark
  Language: it
  NexusMods API key configured: True
  Nexus rate-limit (last response):
    X-RL-Hourly-Remaining: 1997
    X-RL-Hourly-Reset: 2026-07-04 22:00:00 +0000
    X-RL-Daily-Remaining: 19988
    X-RL-Daily-Reset: 2026-07-05 00:00:00 +0000

--- SNAPSHOT ---
  Files tracked: 220
[...]

PR #252, opened 2026-07-05
iteminfo: version-adaptive layout — decode CD 1.13 so Format 3 stack mods apply (#247)

raxjinn:
## What this does

Adds **real CD 1.13 `iteminfo` decode** so Format 3 stack-size mods (e.g. Fat Stacks) **apply** on 1.13 instead of being skipped.

This is complementary to #248: that PR fail-fasts the garbage-count spin and cleanly **skips** iteminfo on an unmodelled layout (stops the hang -- great safety net). This PR goes the next step and actually **decodes** the 1.13 layout, so the common stackable-item mods land byte-exact. Verified the two compose (see Compatibility below).

Closes the iteminfo half of #247 for stackable items.

## What 1.13 changed (reverse-engineered vs the pre-1.13 fixture)

- `_prefabDataList` and `_gimmickVisualPrefabDataList` were **relocated** out of their mid-record slot (right after `_dropDefaultData`) to the **record tail**, and restructured.
- Equipment records have **further** changes beyond that, clustered in the `fixed_page_data_list` / `dynamic_page_data_list` / inspect / socket region (divergence tally: 2488 records diverge at `fixed_page_data_list`, 422 at `dynamic_page_data_list`, then a long tail). Those inner structs (the ones already parsed opaquely pre-1.13) are **not** decoded here.

## Approach -- version-adaptive layout (no hardcoded version check)

- `_ITEM_FIELDS_CD113`: the field order with the two relocated fields dropped; the relocated tail block is carried verbatim via the existing `_tail_slack` mechanism.
- `detect_iteminfo_layout(data, record_offsets)`: samples a handful of `.pabgh`-framed records and picks the layout (`default` vs `cd113`) that round-trips the most records **byte-exact**. Self-selecting -- a future layout that matches neither falls through to the existing per-record opaque carry, so the whole-table round-trip stays byte-exact regardless.
- `_read_item` / `_write_item` / `parse_iteminfo_from_bytes` / `serialize_iteminfo` take an optional `fields` layout (default `_ITEM_FIELDS`, so existing callers are unchanged).
- `iteminfo_writer._build_change_relocated_layout()`: whole-table writer for the relocated layout -- decoded (stackable) records edit via normal dict assignment; records still carried opaque (equipment) take a fixed-offset leading-scalar byte-patch (`max_stack_count`, `is_blocked`); falls back to the pure byte-patch writer if the variant can't round-trip.

## Verification

Against the **live 1.13** `iteminfo.pabgb` (6,508 records):
- **100% byte-exact round-trip** (3,167 records fully decoded + editable; 3,341 carried verbatim). No watchdog stall.
- **Fat Stacks (2,254 `max_stack_count` intents): 2,254/2,254 applied, byte-exact**, 0 stray bytes.
- Pre-1.13 fixture: still selects the `default` layout, byte-exact round-trip unchanged.
- `260` iteminfo/format3 tests pass, **0 regressions**. New: `tests/test_iteminfo_layout.py`.

Repro/verify harness: `scripts/verify_113_parser.py`.

## Compatibility with #248

Applied #248's `_Reader.carray` guard on top of this branch and re-ran the live-1.13 verification: still **100% byte-exact** (the guard never fires on valid decodes; it only fast-fails the equipment records we already carry opaque, which is strictly good). The two changes are orthogonal -- #248 in `carray`, this in the field layout + writer -- and merge cleanly.

## Scope / follow-up

Equipment-only structs (sockets / pages / tribes) reshuffled by 1.13 are still carried verbatim, so equipment-stat mods aren't editable yet -- but nothing corrupts (byte-exact round-trip), and the divergence map above is the starting point to finish full decode.

### DIFF
--- a/src/cdumm/engine/iteminfo_native_parser.py
+++ b/src/cdumm/engine/iteminfo_native_parser.py
@@ -297,6 +297,7 @@ def _find_next_record_start(
 def parse_iteminfo_from_bytes(
     data: bytes,
     record_offsets: "list[int] | None" = None,
+    fields=None,
 ) -> list[dict]:
     """Parse an entire iteminfo.pabgb body to a list of item dicts.
 
@@ -324,7 +325,7 @@ def parse_iteminfo_from_bytes(
                            else len(data))
                 r = _Reader(data, rec_start, rec_end=rec_end)
                 try:
-                    it = _read_item(r)
+                    it = _read_item(r, fields=fields)
                     if r.pos > rec_end:
                         raise ValueError(
                             f"record at 0x{rec_start:X} overran its index "
@@ -363,7 +364,7 @@ def parse_iteminfo_from_bytes(
             next_start = len(data)
         r = _Reader(data, rec_start, rec_end=next_start)
         try:
-            items.append(_read_item(r))
+            items.append(_read_item(r, fields=fields))
             pos = r.pos
         except Exception:
             # GitHub #219: undecodable record (1.12 "*_Flag_I" flags) —
@@ -382,6 +383,7 @@ def parse_iteminfo_from_bytes(
 def serialize_iteminfo(
     items: list[dict],
     offsets_out: dict[int, int] | None = None,
+    fields=None,
 ) -> bytes:
     """Inverse of parse_iteminfo_from_bytes. The byte output must be
     identical to the input when items haven't been modified.
@@ -393,7 +395,7 @@ def serialize_iteminfo(
     for it in items:
         if offsets_out is not None:
             offsets_out[it["key"]] = len(w.buf)
-        _write_item(w, it)
+        _write_item(w, it, fields=fields)
     return bytes(w.buf)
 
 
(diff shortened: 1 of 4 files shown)


AgentKush:
Independently validated this on a live CD **1.13** install and it holds up beautifully -- genuinely impressive reverse-engineering, and thank you for taking it all the way to a real decode instead of just skipping.

**What I ran on my end (not just re-reading the PR):**

- Round-tripped the real `iteminfo.pabgb` (5,938,891 bytes, 6,508 records framed from the `.pabgh` index) through `parse_iteminfo_from_bytes` + `serialize_iteminfo` under `_ITEM_FIELDS_CD113`:
  - **byte-exact -- output identical to input**, 5,938,891 bytes back out
  - 3,167 records fully decoded + editable, 3,341 carried opaque, 3,167 with `_tail_slack` -- matches your reported split exactly
  - ~16s, no watchdog stall
- For contrast, the `default` (pre-1.13) layout on the same 1.13 data is not byte-exact and spins on the opaque-salvage path -- so `detect_iteminfo_layout`'s "pick the layout that round-trips the most records" approach is exactly right: a future drift just falls through to the opaque carry instead of corrupting.
- Merged this branch on top of the current integration tree (which already carries #248's `carray` guard): **clean merge, no conflicts**, orthogonal changes -- exactly as you described in the Compatibility section.

The version-adaptive design (no hardcoded version gate, layout self-selected by byte-exact round-trip) is the right pattern for a game that keeps reshuffling these layouts between patches.

This is a big unlock -- with 1.13 iteminfo decoding byte-exact, stackable-item scalar edits (stack size, block flag, ...) become safe to surface in tooling. It closes a wall I'd personally been stuck behind on the 1.12 -> 1.13 drift.

Big thanks @raxjinn. :pray:

PR #261, opened 2026-07-08
feat(gear-stats): decode + byte-exact edit armor/weapon/AbyssGear stats on 1.13

AgentKush:
Adds the ability to read and edit **gear stats** — armor defense, weapon damage, and the AbyssGear enhancement values — which are currently unmoddable in CDUMM. This is the single most-requested Crimson Desert modding capability and the main reason people reach for DMM.

### The problem
On CD 1.13 the *equipment* iteminfo record tail is a distinct, variable layout the whole-record parser doesn't model, so those **3,341 gear records are carried `_opaque_record`** (byte-exact, but no decoded fields → nothing to edit). The stats live inside `EnchantStatData` blocks in that tail: carrays of `EnchantStatChange {stat: u32, value: i64}`.

### The approach
You don't need to decode the whole record to edit a stat. `gear_stats.py`:

1. **Locates** `EnchantStatData` blocks structurally (four consecutive carrays), and
2. gates each block with a **whitelist of stat keys derived adaptively from the table itself** (keys that recur ≥10× across the opaque records). Real stat keys recur in the thousands; a coincidental byte pattern almost never hits a specific high-frequency key. No game-version constant is baked in — it self-adjusts if the game shifts.
3. **Edits** a value as a **same-width `i64` overwrite** → byte-exact, the record length and companion `.pabgh` index never change. Same safety property as the item-price writer.

### Verification
Live against the installed 1.13 `iteminfo` (6,508 records, 3,341 opaque):

- stats located in **3,313 / 3,341 (99%)** gear records — **21,387** editable stats
- **3,313 / 3,313** same-width edits round-trip **byte-exact** (0 collateral)
- AbyssGear DDD items decode cleanly: LV1 = 1000, LV2 = 2000, LV3 = 3000
- 0 false positives on decoded (non-gear) records under the intended opaque-only scope

**11 CI tests** (`tests/test_gear_stats.py`, synthetic fixtures, no game data) pin the locate/precision/edit invariants.

### Scope
This PR is the **engine** (locate + adaptive whitelist + byte-exact editor) — self-contained, no coupling to the native parser, CI-testable. The natural follow-ups, kept separate for reviewability:

- wire the located stats into the **Format 3 apply path** for opaque iteminfo records (a gear-stat intent → `edit_record_stats` → byte-exact), so `.field.json` gear mods apply; and
- surface them in the **Game Data tab mod maker** as editable cells.

The enchant-struct field names follow the existing `crimson-rs` lineage already in the iteminfo parser (`_read_EnchantStatData` etc.).

---

### Update: now wired end-to-end (applies through Format 3)

Follow-up commit adds the apply integration so gear-stat mods actually work, not just the library:

- `gear_stat[<statkey>]` is the Format 3 field address (e.g. `gear_stat[1000002]`). The number is a **stat key** (first match — "set defense to X") when it's in the stat band, else a positional index.
- `format3_handler` early-accepts `gear_stat[...]` for iteminfo, so mods **validate** (supported, not skipped).
- the iteminfo writer's opaque-record path **locates the stat and overwrites the i64 same-width** → byte-exact; the adaptive whitelist is built lazily only when a gear-stat intent is present.

**Verified live 1.13 end-to-end:** a mod setting `gear_stat[1000002] → 99999` on the AbyssGear DDD item validates (supported = 1) and applies **byte-exact** — 3 bytes changed (the i64 delta), **0 collateral records**, no `.pabgh` rewrite. **280 format3/iteminfo tests pass; 13 gear_stats CI tests.**

So a `.field.json` like this now imports + applies in CDUMM:
```json
{ "format": 3, "targets": [ { "file": "iteminfo.pabgb",
  "intents": [ { "entry": "Item_Stat_AbyssGear_DDD_LV1", "key": 1002785,
                 "field": "gear_stat[1000002]", "op": "set", "new": 99999 } ] } ] }
```

Remaining follow-up: surface located stats in the Game Data tab mod maker as editable cells (GUI only — the engine + apply are done here).

---

### Also in this branch (Game Data tab + packaging — added while iterating)

Beyond the gear-stat engine + apply, this branch also carries the Game Data tab work and the packaging fixes that make the shipped exe correct:

**Game Data tab**
- **Gear-stat editor GUI** — the located stats surface as editable cells that emit `gear_stat[key]` Format 3 intents.
- **Off-thread audio** decode with a live countdown timer (fixes the UI freeze + frozen timer when playing `.wem` audio).
- Help/how-to layout: the modding-help dropdown beside the largest-keyed tables, plus a "how to make a mod from game data" box.

**Packaging / previews (shipped-build correctness)**
- Bundle **Pillow** so DDS texture previews decode in frozen builds.
- `strip=(os.name != 'nt')` — `strip=True` corrupted `python313.dll` on Windows CI (exe would not launch).
- Bundle **Qt3D** + `Qt6Network` + `Qt6ShaderTools.dll` so the texture **3D preview** works when frozen (RHI/DirectX backend); defer the 3D window creation to avoid a COM input-sync crash; hide it where it cannot run.
- Strip the confirmed-unused `opengl32sw.dll` to slim the exe.

### DIFF
--- a/cdumm-macos.spec
+++ b/cdumm-macos.spec
@@ -159,6 +159,10 @@ a = Analysis(
         'cdumm.engine.language',
         'cdumm.engine.compiled_merge',
         'cdumm.engine.texture_mod_handler',
+        # PIL DDS texture decode for the Game Data preview — name the core
+        # codec + DDS plugin so the build actually bundles them (Pillow
+        # registers plugins lazily, which PyInstaller can miss).
+        'PIL', 'PIL.Image', 'PIL.ImageFile', 'PIL.DdsImagePlugin',
         'cdumm.archive.pathc_handler',
         'cdumm.engine.mod_health_check',
         # Imported by the GUI even though the page is hidden on macOS , 
@@ -248,9 +252,10 @@ a = Analysis(
         'PySide6.QtTest', 'PySide6.QtDBus', 'PySide6.QtConcurrent',
         # scipy/numpy, only needed for acrylic blur (disabled)
         'scipy', 'numpy', 'numpy.core', 'numpy.linalg',
-        # PIL/Pillow, not imported by CDUMM (colorthief dep, unused)
-        'PIL', 'PIL._imaging', 'PIL._avif', 'PIL._webp', 'PIL.Image',
-        'Pillow', 'colorthief',
+        # PIL/Pillow IS used — it decodes DDS textures for the Game Data
+        # preview, so it must be bundled (do NOT exclude it). Only colorthief
+        # (which merely pulls PIL in transitively) is unused.
+        'colorthief',
         # brotli, not used by CDUMM (transitive dep from py7zr)
         'brotli', '_brotli', 'brotlicffi',
         # cryptography used by privatebin for AES-GCM + PBKDF2. Keep minimal subset.
--- a/cdumm.spec
+++ b/cdumm.spec
@@ -62,11 +62,22 @@ _qflw_datas, _qflw_binaries, _qflw_hiddenimports = collect_all('qframelesswindow
 # (GitHub #175 / #178 / #179: CERTIFICATE_VERIFY_FAILED).
 _certifi_datas = collect_data_files('certifi')
 
+# Qt6ShaderTools.dll — Qt3D's default RHI render plugin (rhirenderer.dll, the
+# DirectX backend) links it. Without it the plugin fails to load and the
+# texture 3D preview hard-crashes in the frozen build. PySide6's hook doesn't
+# pull it when QtShaderTools isn't imported, so bundle the DLL explicitly next
+# to the other PySide6 Qt DLLs.
+import PySide6 as _pyside6
+_shadertools_binaries = []
+_st_dll = os.path.join(os.path.dirname(_pyside6.__file__), 'Qt6ShaderTools.dll')
+if os.path.isfile(_st_dll):
+    _shadertools_binaries.append((_st_dll, 'PySide6'))
+
 
 a = Analysis(
     ['src/cdumm/main.py'],
     pathex=['src'],
-    binaries=_xxhash_binaries + _native_binaries + _crimson_rs_binaries + _qflw_binaries,
+    binaries=_xxhash_binaries + _native_binaries + _crimson_rs_binaries + _qflw_binaries + _shadertools_binaries,
     datas=[('cdumm.ico', '.'), ('asi_loader/winmm.dll', 'asi_loader'),
            ('src/cdumm/translations', 'cdumm/translations'),
            ('schemas/pabgb_complete_schema.json', 'schemas'),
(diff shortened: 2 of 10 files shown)


AgentKush:
Converting this to a draft — it's been superseded by #277, and I want to flag that clearly before anyone merges both.

**What changed underneath it.** This PR treats equipment records as opaque:

> "those 3,341 records are carried `_opaque_record` (raw bytes, byte-exact but not field-decoded)… we locate the `EnchantStatData` blocks structurally"

That was true when I opened this. It isn't any more. #274 decodes the CD 1.13 equipment records **fully** (0 opaque, 6508/6508, byte-exact), so #277 reads the *real* decoded stat fields instead of heuristically scanning raw bytes for stat-shaped blocks. The heuristic here — an adaptive stat-key frequency whitelist to avoid false positives — exists only because the records didn't decode. Now that they do, it's strictly more fragile than reading the fields directly.

So the **engine** in this PR (`gear_stats.py`) is obsolete. #277 also adds things this PR doesn't: authoritative stat names read from `statusinfo.pabgb` (the community map I was implicitly trusting turned out wrong on 18 of 24 ids), a verification harness, and both path dialects through the shared writer.

**What's still worth keeping here:** the **GUI editor** in `game_data_page.py` (+373). #277 makes gear stats editable via Format 3 `.field.json` mods but has no point-and-click UI. That's the real delta, and it should be re-based to drive #277's native decode rather than this PR's heuristic locator.

**Suggested path:** merge #277 first, then I'll open a small follow-up that ports just the GUI onto the real decode. Leaving this as a draft so the GUI work isn't lost and nobody accidentally merges two competing stat engines into `iteminfo_writer.py` (they don't textually conflict — git would silently take both — which is the trap).

Reopen-ready if you'd rather go a different way. Sorry for the duplicate — I built #277 without spotting this was already open.

AgentKush:
Closing this draft — **every commit in it is already in #269** (rebased, same subjects), so it adds nothing to review. But it shouldn't just be closed quietly, because the gear-stat work it introduced is still sitting inside #269, and I now have measurements saying it's wrong.

### The scanner in here doesn't work

`gear_stats.py` locates stats by **scanning raw record bytes** for plausible-looking keys (`_STAT_KEY_MIN/MAX`, `MIN_STAT_FREQ = 10`, `_VALUE_ABS_MAX`). That was a reasonable thing to write when equipment records were opaque blobs and there was no other way in. Since #274/#277 they aren't — all 3,341 equipment records decode, and the stats have exact addresses.

So I ran the scanner against the real CD 1.13 table and diffed it against the native decode:

| | |
|---|---|
| Ground truth (native decode) | **28,081** stat entries |
| This scanner | **21,420** — silently drops **6,661 stats (24%)** |
| Records it disagrees with the truth on | **714 of 3,313** (22%) |
| Records it **invents** stats for (they have none) | **33** |
| Records it misses entirely | 6 |

Concrete case — item **14706**: the real record carries `CriticalRate +1` on **all 11 enchant tiers**. The scanner finds **zero** of them; the frequency/value heuristics throw small values away. A user opens the editor, sees a stat list that's missing a fifth of the item's stats, edits what's shown, and gets a result that doesn't match. No error, no warning.

There's a second, quieter problem: the dialog dedupes by stat key and its own comment admits an edit "resolves to the **first** located entry with that key". A helm carries the same stat id in ~12 places (base + each enchant tier). So editing "attack" would change the base and silently leave every upgrade tier alone — the number would drift the moment you enhanced the item.

### What happens instead

The GUI itself is fine and worth keeping — it's only the data layer underneath it that was guessing. I'm rebuilding the editor on top of the real decode (`sharpness_data.stat_list[i].change_mb`, `enchant_data_list[t].enchant_stat_data.<list>[i].change_mb`), with names resolved from the game's own `statusinfo` table rather than a hardcoded map, and every tier addressable instead of deduped away.

Superseded by the decode stack (#274 → #277). `gear_stats.py` is being removed from #269 separately so the scanner doesn't ship with the media/preview work.

PR #201, opened 2026-06-09
 feat(tools): enable Find Culprit Mod on non-Windows hosts

RoGreat:
It accomplishes what a Windows host would do by finding the game process (under Proton it's the `S:\` drive pointing at the `Steam\steamapps\common` directory) and polling the process every second with `psutil`. This should be about the same process used before.

Maybe a more advanced version could parse through the logs in `AppData/Local/Pearl Abyss/log`. For example, the error message for not failing to load in game data (had to throw it into a translator) is:
```
(2026-06-08 18:55:36) (0x352 ) [ERROR] [Success](0) [게임데이터 로딩에 실패했습니다.]]
(2026-06-08 18:55:37) (0x352 ) [I] [Success](0) [Terminate 시작.]]
```
This results in a crash before the game gets to the menu. There might be other ways the game crashes on entering the game so maybe it's best not to do that and just wait the full 90 second duration.

Issue #195, opened 2026-06-05
Find culprit mod for unix systems


RoGreat:
Sounds like a cool idea. Although I think it's only on Windows for now so haven't really tried it.

This could be a feature request or maybe disabled similar to the ASI plugins section.

RoGreat:
I have prototyped something to work (off of v3.3.16) . Of course since there aren't any dumps to determine the cause of the program exit I just tuned the wait time until it works. Here's what I changed: https://github.com/RoGreat/CrimsonDesert-UltimateModsManager/compare/master..unix-culprit

Maybe I'll take another look into how to determine the program exit within proton or some other way. But it is not as feature complete as the Windows version at the moment, and I don't think it is portable for MacOS.

PR #262, opened 2026-07-08
feat(tools): enable Find Culprit Mod on non-Windows hosts (continues #201)

AgentKush:
Continues #201 (@RoGreat) — enables **Find Culprit Mod on macOS / Linux**, addressing the review feedback there. Credit to @RoGreat (co-authored).

### Why it can run cross-platform
The bisection classifies each round crash-vs-stable purely from whether `CrimsonDesert.exe` appears and then survives (`find_game_process` + `wait_for_exit`). It does **not** use the Windows-only Pearl Abyss crashpad `.dmp` files in its decision, so it runs anywhere the process can be found. Under Proton the game runs as `…/CrimsonDesert.exe PlatformServiceType=Steam`, so the basename is findable (confirmed against the `ps` tree @RoGreat posted in #201).

### Addresses the review on #201
- **Hard-coded install path → executable basename.** `find_game_process` scans `psutil.process_iter` and matches `CrimsonDesert.exe` by basename — via `ntpath.basename`, since the cmdline path is Windows-style even on Linux — not a literal `S:\…` path, so it works on any machine. Per-process psutil errors and empty/None cmdlines are skipped, so a vanishing/unreadable process can't break the scan, and `crashpad_handler.exe` in the same folder isn't mistaken for the game.
- **Bare `except` in `wait_for_exit`.** Now catches only `psutil.TimeoutExpired` (→ still running) and `psutil.Error` (→ gone); `KeyboardInterrupt` / `SystemExit` propagate.
- **Dropping the #195 guard.** The guard existed because the finder never matched on non-Windows, so every round reported a false crash. With a working finder that root cause is gone, so Start is enabled cross-platform.
- **Leaked cruft** (file rename, issue-template deletion, tr()→hardcoded English) — not present; this is a clean reimplementation off current `master`, so translations for all 15 languages are untouched.

### Tests
Replaces the obsolete `test_find_culprit_windows_only.py` with `test_unix_find_culprit.py` — 11 tests that mock `psutil` and force the non-Windows branch, so they run in CI on Windows too: basename match (case-insensitive, crashpad-handler-not-matched, skip bad procs), `wait_for_exit` running/exited/gone + `KeyboardInterrupt` propagation, and `kill_process` SIGTERM + already-gone guard.

Note: I don't have a Linux/macOS host for a live end-to-end bisection — the logic is unit-tested and the basename match is confirmed against @RoGreat's Proton process tree. A Linux user smoke-test would be the ideal final check.


AgentKush:

PR #265, opened 2026-07-08
fix(crash-report): preserve crash trace across relaunch + surface native faults

AgentKush:
## Summary

The **CRASH TRACE (previous session)** section of the bug report was effectively dead for real crashes and could show a misleading trace. Root cause: `main.py` opens `crash_trace.txt` with `"w"` (truncate) on **every** startup with no preservation, so a hard crash's faulthandler dump is wiped by the very next launch — before the user can generate a report. For native faults (e.g. the Qt3D preview crash) the only trace that ever showed was unrelated Python frames (`splash.py`).

This makes the crash trace actually survive, and readable at a glance.

## Changes

**1. Trace survives relaunch** (`main.py`) — `_preserve_prior_crash_trace()` renames `crash_trace.txt` → `crash_trace.prev.txt` **before** faulthandler truncates it, at startup (AppData + temp fallback). Best-effort, wrapped so it can never block boot. `os.replace` keeps the most recent previous session.

**2. Read the survivor** (`bug_report.py`) — the report reads `crash_trace.prev.txt` first (the real previous session), falling back to the live `crash_trace.txt` for a rare same-session dump. First candidate with content wins.

**3. Headline the crash type + flag native faults** (`bug_report.py`) — faulthandler writes the fault class on line 1 (`Windows fatal exception: access violation`, `Fatal Python error: Segmentation fault`). That headline is surfaced in the **TL;DR** instead of buried. Native faults are labelled so triagers know the Python frames may be unrelated to the real C/C++ crash site.

**4. GPU / Qt render diagnostics** (`bug_report.py`) — the SYSTEM section reports Qt version, QPA platform plugin, `QSG_RHI_BACKEND` override, and GPU adapter name + driver version. Render-stack faults (3D preview, DDS decode) are the most common native crashes. All lookups are read-only — GPU comes from the Windows registry, **no GL context and no subprocess** — so the crash reporter can't crash itself on a bad GPU driver.

## Testing

`tests/test_crash_report.py` — 13 new tests: preservation (move / skip-empty / no-op / keep-most-recent), headline + native detection on real faulthandler samples, candidate ordering (prev before live), safe GPU probe shape, and end-to-end report wiring (a preserved native crash surfaces with headline + note and beats the same-session dump; no trace ⇒ no false "crash detected"). 34 related tests pass locally (new + existing crash / bug-report / i18n-parity). No behavioural change to any non-crash path.

### DIFF
--- a/src/cdumm/gui/bug_report.py
+++ b/src/cdumm/gui/bug_report.py
@@ -385,6 +385,132 @@ def _exe_info(exe_path: Path) -> str:
         return f"{exe_path} (error: {e})"
 
 
+# ── Crash-trace parsing / GPU diagnostics ──────────────────────────────
+
+# faulthandler headline prefixes that indicate a *native* fault: the
+# crash is in C/C++ code and the Python frames it prints may be unrelated
+# to the actual fault. Matched case-insensitively against the first
+# non-empty line of the trace.
+_NATIVE_CRASH_MARKERS = (
+    "windows fatal exception",   # access violation, stack overflow, ...
+    "segmentation fault",
+    "bus error",
+    "illegal instruction",
+    "fatal python error: aborted",
+)
+
+
+def _crash_trace_candidates(app_data_dir: Path) -> list[tuple[str, Path]]:
+    """Return (label, path) crash-trace sources in priority order.
+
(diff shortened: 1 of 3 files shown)


AgentKush:

PR #268, opened 2026-07-09
fix(game-data): don't render packed files' random ASCII as a names outline

AgentKush:
## Problem
On the Game Data tab, packed / binary assets were rendered with fake, misleading "structure," in four distinct ways:

1. **Fake "names" outline on packed value-only files.** Previewing e.g. `gamedata/iteminfo.pabgh` showed *"Structure outline — the field / type / object names embedded in this reflection-serialized binary"* full of random junk (`&N$0`, `_nZnl1`, `;@g#`, `JJ<sR`, `WD":` …). Packed formats (`.pabgh` / `.paatt` / `.paac` / …) store **values only** — they embed no names — but `extract_strings` surfaced *any* printable-ASCII run of >= 4 chars as an "embedded name."
2. **The same fake outline on other packed / third-party formats** — `.hkx` (Havok type tags + FourCC markers), `.roadsector` (`navigraphX` repeats), `.pat`, `.uianiminit`, `.pbd`, `.paac`, `.pam`, `.padxil` — whose incidental ASCII was identifier-shaped enough to slip a pure content filter.
3. **`.mp4` mis-parsed as a "reflection schema."** `decode_reflection` false-positived on video binaries and rendered ~130 garbage "fields" (`_.<yL`, `_aKv(`, `_G"~IA/`), because it accepted any run starting with `_` as a field.
4. **Audio Play flashed a console window** each time `vgmstream` ran as a subprocess.

## Fix
Layered so both the flat *outline* and the rich *schema* views only fire for the game's genuine reflection formats:

- **`extract_strings` name filter** — a run is kept only if it's an engine identifier (`Sequence`, `_isAccessLock`, `bool`) or an asset path (contains `/`). Punctuation-laden noise is dropped.
- **`REFLECTION_EXTS` allow-list gate** (`game_data_page._on_preview_ready` routing) — the flat "names" outline now renders **only** for the verbose reflection formats (`.pae` / `.paseq` / `.prefab` / `.meshinfo` / `.paproj` / `.pastage`). Every other binary routes to the honest typed-word struct / hex view. One rule fixes `.pabgh`, `.hkx`, `.roadsector`, `.pat`, `.uianiminit`, `.pbd`, `.paac`, `.pam`, `.padxil`.
- **`decode_reflection` field validation** — `_field` names must be real identifiers (`_[A-Za-z][A-Za-z0-9_]*`), so `.mp4` and other binaries with punctuation `_`-runs no longer parse as a fake schema.
- **`.pastage`** added to `REFLECTION_EXTS` — it *is* a genuine reflection / quest-stage format (opens as a real field/type table).
- **`convert_to_wav`** now passes `CREATE_NO_WINDOW` on Windows — no more console flash on audio Play / Export.

## Tests (39 pass)
- `test_extract_strings_drops_packed_noise` — punctuation noise -> `[]`; real identifiers + an asset path still pass.
- `test_reflection_exts_gate_membership` — verbose formats IN; packed / third-party (`.hkx` / `.roadsector` / `.pabgh` / `.paatt` / `.paac` / `.pabgb`) OUT.
- `test_decode_reflection_rejects_punctuation_field_noise` — mp4-style `_`-punctuation noise -> `None`; a clean object + `_field`/type stream still parses (>= 3 fields).
- Existing `test_extract_strings_pulls_field_names` stays green.

Verified in a packaged onefile build: every packed / binary format above now shows an honest struct or hex view instead of fake "names," and `.mp4` no longer shows a garbage schema.

### DIFF
--- a/src/cdumm/engine/game_index.py
+++ b/src/cdumm/engine/game_index.py
@@ -27,6 +27,13 @@ from typing import Any, Callable, Iterable
 # Data-table blob (.pabgb) + schema/header (.pabgh) extensions.
 TABLE_EXTS = (".pabgb", ".pabgh")
 
+# The game's own verbose reflection-serialized formats — these embed a real
+# field / type / object name schema as text (readable via decode_reflection /
+# extract_strings). Packed value-only formats (.pabgh/.paatt/.paac/…) and
+# third-party binaries (.hkx, .roadsector, …) embed NO name schema, so the
+# flat "names" outline must never be mined from them.
+REFLECTION_EXTS = (".pae", ".paseq", ".prefab", ".meshinfo", ".paproj")
+
 
 def category_of(path: str) -> str:
     """First path segment, e.g. 'gamedata' for 'gamedata/iteminfo.pabgb'."""
--- a/src/cdumm/engine/game_index.py
+++ b/src/cdumm/engine/game_index.py
@@ -604,11 +604,34 @@ def convert_to_wav(data: bytes, out_wav: str) -> bool:
             pass
 
 
+_NAME_LIKE = re.compile(r"[A-Za-z_][A-Za-z0-9_]*$")   # engine identifier
+
+
+def _is_name_like(s: str) -> bool:
+    """True only for strings shaped like an engine reflection name — a plain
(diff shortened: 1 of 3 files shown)


AgentKush:

PR #270, opened 2026-07-10
fix(game-data): show every indexed file type in the Type filter (drop the top-40 cap)

AgentKush:
## Problem
On the Game Data tab, once the index is built the **Type** filter dropdown is repopulated from the index - but it kept only the **40 most common extensions** (`rows[:40]`, ordered by count). A real index has ~87 distinct extensions, so the rare-but-important ones fell off the list.

Most glaring: **`.pabgb`** - the keyed data-table format the mod maker actually edits - ranks ~48th (~130 files), because bulk assets (`.wem` 375k, `.paa` 316k, `.dds` 283k, ...) crowd it out by count. `.paatt` (rank 45) was cut too.

## Fix
Drop the cap - list **every** indexed extension (still most-common first, each labelled with its file count; blank / `(none)` still skipped). It's a scrollable dropdown, so nothing is hidden.

Note: sorted by count, `.pabgb` lands ~48th, so it's a scroll to reach. Pinning the modding formats to the top is an easy follow-up - kept out of this change to stay minimal.

## Tests
`tests/test_game_data_type_filter.py` drives the real `_populate_type_filter` with a stub `self` against an in-memory index:
- a rare format past position 40 is retained (no cap),
- blank / `(none)` extensions are skipped,
- count labels are comma-grouped.

47 game_data / game_index tests pass.

### DIFF
--- a/src/cdumm/gui/pages/game_data_page.py
+++ b/src/cdumm/gui/pages/game_data_page.py
@@ -1023,8 +1023,13 @@ class GameDataPage(ToolPageBase):
         return exts[idx] if 0 <= idx < len(exts) else None
 
     def _populate_type_filter(self, con) -> None:
-        """Fill the Type dropdown from the extensions actually indexed, most
-        common first, each labelled with its file count."""
+        """Fill the Type dropdown from EVERY extension actually indexed, most
+        common first, each labelled with its file count.
+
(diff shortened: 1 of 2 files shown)


AgentKush:

PR #271, opened 2026-07-10
feat(format3): full patch-op parity , match + clone/delete/new_record/array_append

AgentKush:
> **Requested in #272** (@pinapana 鈥?*Crazy ExtraSockets*). Part of a 3-PR stack that makes that mod work end to end; **#275** is the one that closes it.
>
> `#271` (this) implements the `match` selector 路 **#274** makes CD 1.13 equipment decode at all 路 **#275** routes `match` through the native decoder so it actually selects those records.
>
> Verified against pinapana's real mod once all three are applied: its 3 `match` intents expand to **9,051 concrete edits** and change **3,017 items**, all via `drop_default_data`, with **zero** collateral edits. Before: 0.

---

Brings CDUMM's Format 3 (`.field.json`) engine to full patch-op parity with the wider mod ecosystem 鈥?a `match` selector plus the four record ops (`clone_record`, `delete_record`, `new_record`, `array_append`) 鈥?all built from CDUMM's own engine (no third-party parser/schema) and verified byte-exact against a real vanilla game table.

Every op is safe by construction: the reshaping ops are **append-only or full-rebuild + a mandatory parse-back self-check** (the result is re-decoded and verified before anything is committed; any failure returns `None` and the op is skipped, never applied). All are isolated to the mods that use them 鈥?every existing Format 3 flow is unchanged.

## Ops

**`match` selector** 鈥?one intent edits every record whose fields all equal given values (AND). Expand-to-set: decoded via the existing `parse_records`, each match becomes an ordinary per-record `set` through the trusted writer + verified-field gate.
```json
{ "match": { "_storeType": 3 }, "field": "_buyPriceRate", "new": 50 }
```

**`clone_record`** 鈥?deep-copy a record to a new key/name + patch the copy (the standard item/gear-variant workflow). Append-only + self-checked. Patches target verified scalar fields, and on `iteminfo` also **`gear_stat[...]`** (armor defense / weapon damage) 鈥?so you can clone a weapon and buff its damage on the copy, byte-exact (a real clone-a-weapon round-trip on the live iteminfo table leaves the source + all other records identical). The gear-stat support degrades gracefully (skips with a reason) in a build where the gear-stat feature isn't present, so this PR stays independent of it.
```json
{ "op": "clone_record", "source_key": 1000080, "new_key": 1000090, "new_name": "Hwando+", "patches": [{ "field": "_itemTier", "new": 5 }] }
```

**`delete_record`** 鈥?remove a record by key. Rebuilds the body from the survivors and reindexes the `.pabgh`; self-checked (deleted key gone, every survivor byte-identical, body shrank by exactly the removed entry).
```json
{ "op": "delete_record", "key": 1000090 }
```

**`new_record`** 鈥?build a record from an existing template (`source_key`/`template_key` + `new_key` + `new_name` + `patches`), routed through the same self-checked clone engine. Without a template it's skipped with guidance (a bare field-list build needs a per-table serializer; cloning a working record is the recommended path).

**`array_append`** 鈥?appends one element to a list field, leaving existing elements byte-identical. Live for `dropsetinfo.drops` ("add a loot drop"); gated to fields with a **proven byte-exact list round-trip** (dropset verified on all 14,575 vanilla records). Other list fields get an actionable skip pointing at a full-list `set` until their round-trip is proven.

```json
{ "op": "array_append", "entry": "DropSet_Faction_Graymane", "key": 175001,
  "field": "drops", "new": { "flag": 1, "item_key": 1000080, "rates": 5000 } }
```

The upshot: every DMM-style patch op is handled 鈥?the implementable ops apply byte-exact, the rest skip with a clear, actionable reason. Mod-author reference: `docs/FORMAT3_OPS.md`.

## Verification

- **Unit:** new tests across all ops (byte-exact match鈫攈and-authored equivalence; clone/delete/new_record faithful-copy, self-check refusals, index integrity, composition; array_append validation; validation gates). Full `test_format3_*` + `test_whole_table_*` suites green (**279 passed, 2 skipped**).
- **Real tables:** extracted genuine vanilla tables from an installed copy of the game and ran the ops through the real engine and the real orchestrator (`parse_format3_mod_targets` 鈫?`expand_format3_into_aggregated`): on `wantedinfo` (35 records, u16 keys) match edited exactly the matching records byte-identically to hand-authored sets, clone/new_record produced a faithful copy + patch with all originals intact, delete removed a middle record with survivors byte-identical and the index reshifted, and a real `.field.json` mod file composed match + a record op into the correct whole-table body + `.pabgh` companion; on `dropsetinfo` (14,575 records) parse鈫抯erialize round-trips byte-exact on every record and array_append preserved existing drops byte-for-byte on 200/200 sampled records.

## Provenance

`parse_records`, the existing per-table writers, and the `verified_fields` gate 鈥?CDUMM's own RE, no third-party parser or schema. Self-contained on `upstream/master` (no dependency on the game-data PRs), so it merges in any order.

### DIFF
--- a/docs/FORMAT3_OPS.md
+++ b/docs/FORMAT3_OPS.md
@@ -63,7 +63,22 @@ gear variant.
 - `new_key` must be unused and fit the table's key width; a collision is
   refused.
 - The copy is byte-identical to the source except your patches.
-- Patches target **verified scalar fields** (same rule as `set`).
+- Patches target **verified scalar fields** (same rule as `set`), and on
+  `iteminfo` also **`gear_stat[...]`** — so you can clone a weapon or piece
+  of armour and change its damage/defense on the copy:
+
+```json
+{
+  "op": "clone_record",
+  "source_key": 50003,
+  "new_key": 50003001,
+  "new_name": "Sharpened Blade",
+  "patches": [ { "field": "gear_stat[1000000]", "new": 999999 } ]
+}
+```
+
+  A `gear_stat[...]` edit is a byte-exact same-width overwrite of the stat
+  in the copy; editing a stat the item doesn't carry is a clean no-op.
 
 ---
 
--- a/docs/FORMAT3_OPS.md
+++ b/docs/FORMAT3_OPS.md
@@ -1,0 +1,143 @@
+# Format 3 (`.field.json`) operations — mod author reference
+
+CDUMM applies Format 3 mods by name (entry + field), not raw byte offsets.
+A mod declares a `target` (the `.pabgb` table it edits) and a list of
+`intents`. This document covers every operation CDUMM supports and its
+exact JSON shape.
+
+Every reshaping operation is **safe by construction**: it is append-only or
+a full rebuild, and CDUMM re-decodes the result and verifies it before
+committing. If an operation can't be applied safely it is **skipped with a
+reason** (shown after apply) — it never corrupts the table.
+
+Fields you edit must be **verified** for that table (CDUMM proves a field's
+byte offset against real record data before allowing writes). Unverified
+fields render as `(unverified)` in the Game Data tab and are refused.
+
+---
+
+## `set` — change a field
+
+The default. Change one field on one record, located by `entry` name (and
+optionally numeric `key`).
+
+```json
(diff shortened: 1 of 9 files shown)


AgentKush:

PR #269, opened 2026-07-10
feat(game-data): inline .mp4 video + .bnk playback, scrollbar + Qt3D/multimedia packaging

AgentKush:
Game Data preview enhancements. **Stacks on #261** (gear-stats) for the Qt3D packaging base — merge that first.

## Features
- **In-app `.mp4` video player** — UI "advice" clips play inline via Qt Multimedia (FFmpeg backend), with Pause/Play + a seek slider. Extracts to a temp file, stops + cleans up on switch, and degrades gracefully (with the exact error) if a build lacks multimedia.
- **`.bnk` SoundBank playback** — Play / Export-to-WAV enabled for banks that carry embedded audio (vgmstream decodes the first sound).
- **Scrollbar gutter** — the asset list + preview grid reserve the scrollbar's width so it no longer overlays the last column (mis-clicks).
- **File-type guide** — documents `.mp4` video + `.pastage` / `.pat` / `.pbd` / `.uianiminit`.

## Decode correctness (also in #268)
- Reflection-format **allow-list gate** + tightened `_field` validation, so packed/binary files (`.hkx`, `.pat`, `.pam`, `.mp4`, `.roadsector`, `.uianiminit`, …) stop rendering fake "reflection schema / names" output and fall through to the honest struct/hex view. `+ .pastage` recognised as a real reflection format. Regression tests included.
- `vgmstream` decode now runs with `CREATE_NO_WINDOW` (no console flash on Play).

## Packaging (makes the PUBLIC CI build ship working 3D + video)
- Bundle the Qt3D native DLLs (`Qt63D*`, `Qt6ShaderTools`) and keep `Qt6Network` (Qt63DCore links it) — fixes "DLL load failed importing Qt3DExtras".
- Keep `QtMultimedia` / `QtMultimediaWidgets` / `QtNetwork` + bundle `Qt6SpatialAudio` / `Qt6OpenGLWidgets` + the FFmpeg `av*`/`sw*` codecs — required for `.mp4` playback.

All green (39 tests). Verified in a packaged onefile build: 3D preview loads and `.mp4` clips play with audio.

Issue #246, opened 2026-07-04
[Crash]: Freezes when load gets to 55% and requires Task Manager to close the app


Mourningstar6-6-6:
### When does it crash?

CDUMM crashes on startup

### CDUMM bug report (includes crash trace)

```text
--- SEVERITY: Crash (app closed/froze) ---

============================================================
CRIMSON DESERT ULTIMATE MODS MANAGER -- BUG REPORT
============================================================
Generated: 2026-07-04 16:01:44
App Version: 3.4.2

--- TL;DR ---
  1 issue(s) detected:
    1. 1 mod(s) have source_path pointing to a deleted file — preset re-picking won't work for those.
  OK:
    - Not running as admin (good)
    - All mod-state health checks passed

--- SYSTEM ---
OS: Windows-11-10.0.26200-SP0
Python: 3.13.12 (frozen)
Running as administrator: no
Game Dir: D:\SteamLibrary\steamapps\common\Crimson Desert (exists)
Game Platform: Steam
Game Exe: D:\SteamLibrary\steamapps\common\Crimson Desert\bin64\CrimsonDesert.exe (345.7 MB, mtime 2026-07-04 14:20)
Game Exe Compat Flags: none
Game Version Hash: c5573f24bf24e826
Game Drive Free: 385.8 GB free of 1863.0 GB
App Data: C:\Users\ahdra\AppData\Local\cdumm
App Data Size: 15.3 MB
App Data Drive Free: 92.2 GB free of 930.5 GB

--- STORAGE ---
CDMods: D:\SteamLibrary\steamapps\common\Crimson Desert\CDMods
  vanilla/: 55 files, 3.12 GB
  deltas/: 17 files, 31.9 MB
  sources/: 20 files, 7.1 MB
  overlay/: (not present)

--- SETTINGS ---
  Theme: dark
  Language: (default)
  NexusMods API key configured: True
  Nexus rate-limit (last response):
    X-RL-Hourly-Remaining: 1995
    X-RL-Hourly-Reset: 2026-07-04 21:00:00 +0000
    X-RL-Daily-Remaining: 19952
    X-RL-Daily-Reset: 2026-07-05 00:00:00 +0000

--- SNAPSHOT ---
  Files tracked: 214
  Created: 2026-07-04 19:13:34

--- VERIFY STATE (size-only, fast check) ---
  Tracked: 214
  Match snapshot (likely vanilla): 214
[...]

Issue #236, opened 2026-06-30
FAQ — read this first (mods not applying, install fails, crashes, updates)


AgentKush:
﻿# CDUMM — Frequently Asked Questions

A quick-reference for the questions that come up most often. **Please skim this before opening an issue** — most "a mod won''t work" reports turn out to be one of the entries below.

> 📌 _Intended to be pinned — maintainer, feel free to pin/edit/expand._

---

## Installing & applying mods

**Q. I installed a mod but nothing changed in-game — or only part of it worked.**
Usually the mod was built for an *older game version*. CDUMM edits specific byte locations in the game files; when the game updates, those locations move or change, so a patch built for the old version no longer lines up. CDUMM **skips** any patch whose target bytes don''t match — writing it anyway would corrupt the file. After Apply, CDUMM shows a warning listing exactly which patches were skipped. The real fix is a mod update from its author for your game version; it isn''t something CDUMM can auto-relocate.

**Q. After Apply I get a pop-up: "N patches skipped … expected … got …". Is CDUMM broken?**
No — that''s the safety check working. Those patches expected certain bytes that aren''t present in your game build, so CDUMM declined to write them rather than risk corrupting the file. The mod needs updating for your game version. (If another manager "applies" the same mod without complaint, it''s likely overwriting bytes blindly, which can corrupt the file or crash the effect.)

**Q. Drag-and-drop or install fails (WinError 740 / "access denied" / nothing happens).**
Three usual causes, all outside CDUMM:
1. **CDUMM is running as administrator.** Close it and reopen *without* "Run as admin" — admin breaks drag-and-drop.
2. **The game exe has the "Run as administrator" compatibility flag.** Right-click `CrimsonDesert.exe` → Properties → Compatibility → uncheck "Run as administrator."
3. **The game is installed under `Program Files`.** Windows restricts writes there; move the Steam library to a non-protected drive.
CDUMM''s built-in bug report flags all three in its TL;DR.

**Q. Two mods edit the same file — which one wins?**
The one with the higher priority in your load order. Re-order if the wrong one is winning; the bug report''s "File overlaps" section shows which mods share a file.

**Q. A mod has several folders (body types / variants) — which do I pick?**
CDUMM shows a variant picker on import. Pick the one you want; each variant is a separate sub-mod.

## Updates & game patches

**Q. My mod never shows an available update in the update check.**
[...]

PR #273, opened 2026-07-11
Fix phantom "previous session crashed" in every bug report (and preserve the real crash trace)

AgentKush:
The splash triggers a *continuable* Windows COM warning (`0x8001010d` / `RPC_E_CANTCALLOUT_ININPUTSYNCCALL`) the first time the frameless, translucent splash composits. CDUMM catches it and starts normally — but `faulthandler` still dumps it, so **`crash_trace.txt` is non-empty on every single launch**. Two bugs fall out of that.

### 1. Every bug report headlined a crash that never happened

The crash-trace section flagged *any* non-empty trace, so every report led with "Previous session crashed" plus a `Windows fatal exception` dump. Users and maintainers get sent chasing a phantom.

### 2. The previous session's real trace never survived

`faulthandler` opens `crash_trace.txt` with `"w"` — which **truncates**. The prior run's native trace, which is exactly what a crash report needs, was destroyed at the very next launch, and the report only ever showed the *current* run's harmless splash warning. A genuine crash was already unrecoverable by the time the user could click "generate bug report".

## Fixes

- **`main.py`** moves the prior `crash_trace.txt` to `crash_trace.prev.txt` *before* faulthandler truncates it. Moved (not copied), so a healthy launch can't resurrect a stale "previous crash". A file locked by another instance degrades gracefully (`except OSError`) and never breaks startup.
- **`bug_report.py`** reads the preserved prev trace (falling back to the live file) through a new `classify_crash_trace()` returning `empty` / `benign` / `crash`:
  - a **splash-only `0x8001010d`** dump is **benign** — a quiet OK note, no red flag, no scary dump;
  - a real fault, a benign dump with a real fault **stacked after it**, or an unrecognised dump all still **headline as a crash** (fail safe).

## Verification

10 new tests plus the existing crash / lock / emergency-dump suites (**32 passing**): classifier cases (empty, benign, real access violation, `Fatal Python error`, benign+real stacked, same code raised outside the splash, unrecognised), and `generate_bug_report` integration — including that a **preserved real trace wins over the current benign one**.

Self-contained on `master` (no dependency on the other open PRs), so it merges in any order.

### DIFF
--- a/src/cdumm/gui/bug_report.py
+++ b/src/cdumm/gui/bug_report.py
@@ -177,6 +177,46 @@ def _is_relevant_log_line(ln: str) -> bool:
     return True
 
 
+def classify_crash_trace(text: str) -> str:
+    """Classify a faulthandler ``crash_trace`` dump.
+
+    Returns one of:
+
+    * ``"empty"``  — nothing was recorded.
+    * ``"benign"`` — the ONLY fault recorded is the known Windows splash
+      COM warning ``0x8001010d`` (``RPC_E_CANTCALLOUT_ININPUTSYNCCALL``)
+      raised inside ``gui/splash.py`` during ``show_splash``. Qt/DWM
+      raises this *continuable* exception the first time the frameless,
+      translucent splash composits; CDUMM catches it and starts
+      normally, but ``faulthandler`` still dumps it. It is NOT a crash
+      and must never be surfaced as one, or every bug report headlines a
+      phantom "previous session crashed".
+    * ``"crash"`` — anything else: a real fault, a benign fault with a
+      second fault stacked after it, or an unrecognised dump (fail safe
+      toward "crash" so genuine problems are always surfaced).
+    """
+    if not text or not text.strip():
+        return "empty"
+    # A hard fault is often reported as "Fatal Python error: ..." — that is
+    # never the benign splash case.
+    if "Fatal Python error" in text:
+        return "crash"
+    headers = [ln for ln in text.splitlines()
+               if "Windows fatal exception" in ln]
+    # Benign is exactly one exception dump. Zero (unknown format) or more
+    # than one (a real fault stacked after the splash warning) are not
+    # benign.
+    if len(headers) != 1:
+        return "crash"
+    if "0x8001010d" not in headers[0]:
+        return "crash"
+    # A single 0x8001010d dump is benign only when the crashing frame is
+    # the splash. If any other module is implicated, treat it as real.
+    if "show_splash" in text and "splash.py" in text:
+        return "benign"
+    return "crash"
+
+
 def _format_import_date(d: str | None) -> str:
     if not d:
         return "?"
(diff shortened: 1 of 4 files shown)


AgentKush:
Heads-up for whoever merges: this PR and #265 both fix the "phantom previous-session crash" / crash-trace-preservation problem, and they textually conflict in two files. It's the **only** conflict left across the open PR set (I checked — everything else merges clean, and I already rebased #269 to clear its `game_index.py` clash with #268). It's a small, well-understood resolution, not a rework — here's the exact recipe so it's a 2-minute merge:

**Merge order:** #265 first, then this PR.

**`src/cdumm/main.py`** (1 hunk): both implement "move the prior `crash_trace.txt` to `crash_trace.prev.txt` before faulthandler truncates it." Keep **#265's** version — the `_preserve_prior_crash_trace(...)` helper — and drop this PR's inline `os.replace(...)`. They're equivalent; #265's is the tidier abstraction.

**`src/cdumm/gui/bug_report.py`** (1 hunk): take **both concerns**, not one side:
- keep **#265's** `_crash_trace_candidates()` walk + `_parse_crash_headline()` (the "first trace with content wins; flag native faults" logic), **and**
- fold in **this PR's** `classify_crash_trace()` gate so a benign splash-only `0x8001010d` dump is classified out and never headlines as a crash (a `kind == "crash"` check before emitting the section + TL;DR flag; benign → a quiet note and `continue`).

That combined form is already proven — it's exactly what I resolved for the `release/community-build` testbuild, where #265 + #273 ship together and crash reports come out correct (real hard crashes surface, benign splash dumps don't).

I deliberately didn't force a rebase resolution here: it's 91 lines of crash-reporting logic, and getting it subtly wrong would re-introduce the phantom-crash flag this PR exists to kill. Happy to do the careful rebase in a fresh pass if you'd prefer it pre-resolved — just say the word.

AgentKush:
**Resolved against #265 — merge #265 first, then this.** No force-push; the resolution is a merge commit on this branch.

These two were written in parallel against the same two files (`bug_report.py`, `main.py`) and conflicted head-on. The important part: **neither is correct on its own.**

- **#265 alone re-introduces the phantom crash.** It gates the crash-trace section on the file being *non-empty* — and the benign Qt/DWM splash dump (`0x8001010d`, a *continuable* exception CDUMM catches and starts normally through) **is** non-empty. Every Windows user would get "Crash detected" headlining a perfectly healthy launch. That is exactly the bug this PR exists to remove.
- **This PR alone loses #265's native-fault detection** — which matters. #246 (freeze at 55% load) is very likely a native fault, and without that note the Python frames faulthandler prints actively mislead whoever reads the report.
- **A third hole, in #265's `_preserve_prior_crash_trace()`:** it only calls `os.replace()` and swallows `OSError`. On Windows that rename fails (WinError 32) whenever another process still holds the file — a hung prior instance whose faulthandler handle never closed, or an AV scanner mid-scan — and the `open("w")` right after it then truncates the file **anyway**. So the previous session's trace is destroyed in precisely the hard-crash case we most need it. This PR's `copyfile` fallback is now inside #265's helper.

### The resolution

- `main.py` — keeps #265's extracted helper, with this PR's locked-file `copyfile` fallback inside it.
- `bug_report.py` — keeps #265's candidate walk, headline parsing and native-fault note, but runs `classify_crash_trace()` **first**. A benign dump is reported as an OK observation and its trace isn't dumped; only a real crash reaches the richer reporting.

### Tests

Both suites pass on the result — **24 tests**. Two of this PR's tests pinned the literal flag string `"Previous session crashed"`, which #265 reworded to name the source and the fault; they now assert through a `CRASH_FLAG` constant, same intent.

Added `test_benign_trace_is_not_a_crash_even_with_the_richer_reporting` — the guard for exactly this interaction: a benign dump must produce no crash flag, no CRASH TRACE section, and no native-fault note. Neither PR could have had that test on its own.

Full merge simulation of all 20 open PRs: **0 conflicts** in the order posted on #236.

AgentKush:

AgentKush:

Issue #272, opened 2026-07-10
[Feature]: support for the  new "match" feature in v3 json mods


pinapana:
### Area

None

### What problem would this solve?

Matching huge numbers of keys at once by targeting a single field that is shared across all entries. Brings long v3 mods down to a few, more future-proof/robust intents. Was added to DMM this week at request. My current mods that use this look like this:
[CrazyExtraSockets_v3.1.json](https://github.com/user-attachments/files/29909638/CrazyExtraSockets_v3.1.json)

### Proposed solution

From the new DMM guide, points 08 & 09: 
[V3_MOD_GUIDE.html](https://github.com/user-attachments/files/29909537/V3_MOD_GUIDE.html)

looking for CDUMM to support this syntax and the logging for it:
{
  "match": { "equip_type_info": 2667368044 },
  "op": "set",
  "field": "drop_default_data.socket_valid_count",
  "new": 5
}

### Alternatives or workarounds

Individual key targeting, and using separate mods for CDUMM vs DMM due to differing targeting strategies. Much more verbose CDUMM mods than DMM for mods with many targets. CDUMM mods will deviate from DMM in capabilities.

AgentKush:
Thanks for this — and genuinely thank you for attaching the actual mod file. That is what made this diagnosable, because I could run **your exact mod against the real vanilla `iteminfo` table** instead of guessing from the syntax. That saved a lot of time, and it turned up something important.

## The good news

**`match` already exists** — it's built and open in #271 (batch edits: one intent targets every record whose fields all equal the given values). Reading your mod, I saw it needs one thing #271 didn't have: your selector uses a **list** of 63 `equip_type_info` values, and `match` only did scalar equality. So I've just extended it — a list now means **any-of** (SQL `IN`):

```json
{ "match": { "equip_type_info": [1086980073, 1205223146, ...] },
  "op": "set", "field": "drop_default_data.socket_valid_count", "new": 5 }
```

That's pushed to #271. And the nested `drop_default_data.*` write path already works in CDUMM today (floozo's cloak mod uses `drop_default_data.use_socket` and `.add_socket_material_item_list`), so the syntax you want is supported end to end.

## The honest blocker

I then ran your mod against the real table, and it still selects **nothing**. Here's exactly why, from the live vanilla `iteminfo` (6,508 records):

- `equip_type_info` decodes to only **two** distinct values across the entire table: `None` on **3,341** records, and `0` on the other 3,167.
[...]

PR #274, opened 2026-07-11
iteminfo: decode CD 1.13 equipment (SubItem tag 17 + _enchantDataList) - 0 -> 3151 records

AgentKush:
> **Part of #272** (@pinapana 閳?*Crazy ExtraSockets*). Stack: **#271** (`match` selector) 閳?**#274** (this 閳?makes CD 1.13 equipment decode) 閳?**#275** (routes `match` through the native decoder; the closing keyword lives on that one).
>
> Without this PR the socket mod has nothing to edit: every equipment item was carried opaque, so `drop_default_data` didn't exist to write to.
>
> **Verified with the real mod** (all three applied): 3 `match` intents 閳?**9,051 concrete edits**, **3,017 items changed**, all via `drop_default_data`, **zero** collateral edits, and the patched table still decodes 6,508/6,508 with a correctly rebuilt `.pabgh`. Before: the mod applied cleanly and changed nothing.

---

## The bug

Every equipment record in the live **CD 1.13** `iteminfo.pabgb` failed to decode and was carried opaque 閳?**3151 of 6508 records** with all their fields dropped.

That is what blocked socket editing (#272), gear stats, and left `match` on iteminfo blind to equipment.

## Root cause

Two coupled wire changes that are only correct **together**:

* `SubItem` tag **17** is a *None* sentinel 閳?it carries **no** u32 payload.
* `ItemInfo` grew **`_enchantDataList`** immediately after `_dropDefaultData`: a u32 count, then one `EnchantData` per enchant tier (levels `0..N-1`; 11 tiers on most equipment, 0 on non-equipment).

The old reader spends its 7 bytes as

```
tag(1) + u32 value(4) + svc(1) + use(1)
```

where the truth is

```
tag(1) + svc(1) + use(1) + u32 enchant count(4)
```

**Identical width.** That is why non-equipment 閳?whose enchant count is `0`, so those four bytes are zero either way 閳?round-tripped byte-exact and *never looked broken*, while equipment (count is almost always 11) desynced from `drop_default_data` onward.

It is also why fixing the sentinel *on its own* destroyed non-equipment in earlier testing: that drops the u32 without putting the enchant count back. **Neither change is correct alone**, which is what made this so slippery.

`EnchantData` itself turned out to be exactly the shape the existing (dead, never-wired) `_read_EnchantData` already modelled, plus the u32 that CD 1.12 added to it 閳?so this wires up the existing reader rather than inventing a new struct.

## Results 閳?live 1.13 table, 6508 records

|  | non-equipment | **equipment** | opaque |
|---|---|---|---|
| before | 3167 | **0** | 3341 |
| after | 3357 | **3151** | **0** |

* `serialize_iteminfo()` reproduces the file **byte-for-byte**.
* **804** items now decode with `use_socket = 1` 閳?the #272 targets.
* Decode + re-serialize of the whole table runs in ~1 s.

## Why byte-exactness alone isn't enough

A **wrong-but-same-size** read round-trips byte-exact and would pass every check 閳?while silently mis-assigning fields and corrupting item tables on *write*. So the enchant tier **level ladder** (`0,1,2,閳?N-1`) is asserted as an independent oracle:

* **0 violations** across all 1047 multi-tier records
* the integration test enforces it on **every** record

## Compatibility

Pre-1.13 readers are left untouched. The new shape ships as a separate `cd113_enchant` layout, so `detect_iteminfo_layout` still self-selects the right one per game version.

## Two supporting fixes the new layout depends on

* **`_Reader.carray`** bounded its count by the whole 5.9 MB *body* rather than the *record*, so a desynced count allocated ~500k dicts before failing 閳?a stall, not an error. Now bounded by `rec_end`: a count can never exceed the bytes left in its own record.
* **`detect_iteminfo_layout`** broke ties toward the *first* layout. A sample that happened to draw only non-equipment records would tie `cd113_enchant` with the enchant-blind layout and lock in the latter 閳?silently carrying all 3151 equipment records opaque. Ties now go to the more specific layout; a layout is still never *claimed* when nothing round-trips.

## Tests

`tests/test_iteminfo_cd113_enchant.py` 閳?9 tests, all passing:

* tag 17 consumes no u32; a real tag still carries its u32
* sockets (`add_socket_material_item_list`, `socket_valid_count`, `use_socket`) survive the new DDD 閳?the #272 path
* `EnchantData` round-trips and carries the trailing u32
* **element sizes pinned numerically** (12B / 5B / 20B / 8B) 閳?the one thing a byte-exact round-trip cannot catch
* the `carray` guard bounds by record, not body
* layout tie-break, and "claim nothing when nothing round-trips"
* live-table integration: 0 opaque, level-ladder oracle on every record, byte-exact re-serialize (skips cleanly when the game isn't installed)

Full engine suite: **171 passed, 0 new failures.**

<sub>The 2 remaining failures in `test_iteminfo_walk_real_game.py` are pre-existing on 1.13 閳?they live in the legacy *semantic* walker, reproduce identically on a clean tree, and are untouched by this PR.</sub>

### DIFF
--- a/src/cdumm/engine/iteminfo_native_parser.py
+++ b/src/cdumm/engine/iteminfo_native_parser.py
@@ -196,7 +196,14 @@ class _Reader:
         # parse-guard turns it into a clean "this game version isn't
         # supported yet" skip instead of a hang. Never fires on valid data
         # (a legitimate count is always <= remaining bytes).
-        remaining = len(self.data) - self.pos
+        # Bound by the *record*, not the whole body, when the boundary is
+        # known. A count can never exceed the bytes left in its own record,
+        # so this rejects a desynced count immediately instead of letting
+        # it allocate a list of ~500k dicts first (the body bound is 5.9MB,
+        # which is far too loose to catch anything in time). Matters most
+        # while detect_iteminfo_layout is speculatively trying layouts that
+        # do not fit: without this, every rejected candidate stalls.
+        remaining = (self.rec_end or len(self.data)) - self.pos
         if n > remaining:
             raise ValueError(
                 f"iteminfo carray count {n} exceeds {remaining} remaining "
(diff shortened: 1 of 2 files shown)


AgentKush:
Closing this as **contained in #275** — and it's worth saying why, because merging it separately would have cost you an afternoon.

`feat/iteminfo-native-match` (#275) carries this exact commit, rebased (same subject: *"iteminfo: decode CD 1.13 equipment (SubItem tag 17 + _enchantDataList)"*), and every PR above it in the chain carries it too. So the work isn't lost by closing this — it's already in the thing you'd merge next.

**The problem is that the rebase gave it a different SHA.** I simulated the full merge of all 21 open PRs into a clean `master`, and this is what happens if #274 goes in on its own first:

```
#274  clean
...
#275  CONFLICT -> tests/test_iteminfo_cd113_enchant.py
#277  CONFLICT -> tests/test_iteminfo_cd113_enchant.py
#278  CONFLICT -> tests/test_iteminfo_cd113_enchant.py
#279  CONFLICT -> tests/test_iteminfo_cd113_enchant.py
#280  CONFLICT -> tests/fixture_loaders.py, tests/test_iteminfo_cd113_enchant.py
#281  CONFLICT -> cdumm.spec, tests/fixture_loaders.py, ...
#282  CONFLICT -> tests/test_iteminfo_cd113_enchant.py
#283  CONFLICT -> cdumm.spec, tests/fixture_loaders.py, ...
```

Merging #274 first makes **eight** downstream PRs conflict, on a file whose content is identical in substance. Skip it and the same 19 PRs merge **clean, in order, with no conflicts at all** — I ran it.

Nothing is lost:
- `iteminfo_native_parser.py` is **byte-identical** between this branch and #275.
- `test_iteminfo_cd113_enchant.py` has the **same 9 tests** in #275, but reads from the committed `vanilla113` fixture instead of an env-var path — so they actually run in CI rather than skipping.

Merge #275 instead; it includes all of this. Full merge-order note going on the tracking issue.

PR #275, opened 2026-07-11
format3: fix `match` on iteminfo (0 -> 366 records), nested dotted paths, + camelCase/snake_case parity

### LINKED ISSUE #272: [Feature]: support for the  new "match" feature in v3 json mods

AgentKush:
> **Closes #272** (@pinapana 鈥?*Crazy ExtraSockets*).
>
> Last of a 3-PR stack: **#271** (`match` selector) 鈫?**#274** (CD 1.13 equipment decode) 鈫?**#275** (this). All three are needed; merging this one last completes the feature, which is why the closing keyword lives here.
>
> **Verified against pinapana's actual mod file** (attached to #272), applied to a live vanilla 1.13 `iteminfo` table:
>
> | | before | after |
> |---|---|---|
> | intents its 3 `match` selectors expand to | **0** | **9,051** |
> | items actually changed | **0** | **3,017** |
> | items changed in any *unintended* way | 鈥?| **0** |
>
> Sockets land exactly as authored (`socket_valid_count = 5`, `use_socket = 1`, 5 material rows), and the patched table still decodes 6,508/6,508 records with the enchant-tier oracle intact and a correctly rebuilt `.pabgh` index 鈥?no records lost, none added.
>
> Also fixes the same silent-no-op class for **#191**-style camelCase (DMM-dialect) field names, which resolved for the *write* but not for the *selector*.

---

> **Stacks on #271** (`match` selector) and **#274** (native CD 1.13 equipment decode).

Three commits. All three fix the same failure class: **the mod applies cleanly, reports no error, logs no warning 鈥?and changes nothing.**

---

# 1. `match` on iteminfo silently selected **zero** records

pinapana's Crazy ExtraSockets (#272) is exactly this shape. Two independent causes, either enough on its own:

**Wrong decoder.** `_expand_match_intents` decoded with the generic `parse_records` walker, which only reaches a handful of iteminfo fields. Everything past them 鈥?including `equip_type_info`, which is precisely what the socket mods select on 鈥?came back `None`, so every comparison failed.

**Wrong table name.** `_table_name_from_target` only strips the *extension*, so a path-shaped target yields `gamedata/binary__/client/bin/iteminfo`, not `iteminfo`. That matched no schema, so `parse_records` returned `{}` and the match expanded to nothing **regardless of the decoder**. (`identify_table_from_path` is no help 鈥?it wants the extension that was already stripped.)

**Fix:** route iteminfo through `iteminfo_native_parser` (**116 fields**, not ~5), and take the basename.

Live 1.13 table: a `match` on one real `equip_type_info` now expands to **366** concrete `set` intents 鈥?**was 0**.

---

# 2. `match` selectors can address nested fields

```json
{"match": {"drop_default_data.use_socket": 1}, "field": "...", "new": "..."}
```

Selectors could only compare top-level scalars, so everything interesting about equipment 鈥?sockets, enchant tiers 鈥?was unreachable even though the parser decodes it all. Selecting **every socketed item in one intent** now works: **804** records on the live table.

Lists take an integer index (negatives count from the end):

```
drop_default_data.add_socket_material_item_list.0.item
enchant_data_list.-1.level
```

**No "any element" list traversal 鈥?deliberately.** It's ambiguous against `_match_value_equals`, where a list on the record side already means "match this list exactly". So `some_list.field == 5` could mean "any element has field 5" *or* "the extracted values equal 5"; silently picking one would make a mod select records its author never intended.

---

# 3. `match` keys accept every spelling `field` keys accept

A `match` key resolved snake_case 鈫?camelCase but **not the reverse**, while the writer's `field` key resolves camelCase 鈫?snake_case. So a mod spelling a name one way got a **half-working intent**: the name resolved for the *write* but not for the *selector*, the selector matched nothing, and the intent applied to nothing. This bites **DMM-dialect (v3.1) mods, which author camelCase**.

Both directions now resolve, flat and nested:

```
equipTypeInfo        _equipTypeInfo        dropDefaultData.useSocket
equip_type_info      _equip_type_info      drop_default_data.use_socket
```

The camelCase 鈫?snake_case direction **delegates to the writer's `_resolve_field_name`** 鈥?the *same function*, not a copy, so selector and writer cannot drift apart. That also inherits its separator-insensitive fallback for cases where camelCase word boundaries don't line up with the parser's snake_case (`_equipAbleHash` 鈫?`equipable_hash`, found via pinapana's AbyssGearUnlock report, **#191**).

Resolution order is **exact name 鈫?precise camel/snake transforms 鈫?fuzzy normalised form**, and the fuzzy step accepts only an unambiguous single match, so it refuses rather than guesses. (`iteminfo_writer` doesn't import `format3_apply`, so no cycle.)

---

## Safety

* **Degrades, never guesses.** A native decode that raises, or yields no records, falls back to the generic walker rather than dropping the match.
* **Opaque records are excluded** from the match set 鈥?comparing against stripped fields would silently select nothing.
* **Missing 鈫?no match, never an exception.** A wrong path in a mod cannot break apply.
* **`None` 鈮?missing.** Lookup distinguishes them (`_MISSING`), so a field that genuinely decodes to `None` 鈥?e.g. `default_sub_item.value` on the tag-17 sentinel 鈥?can't fall through to path traversal and pick up something else.
* **A flat field wins over path traversal**, so a field whose real name contains a dot still resolves first.
* **iteminfo only.** Every other table still goes through `parse_records`, with the string handed to it unchanged.
* **No new write path.** Matched records still emit ordinary per-record `set` intents through the existing, already-trusted writer (and its verified-field gate).

## Tests

`test_format3_match_iteminfo_native.py` (7) + `test_format3_match_nested_path.py` (17):

* routing: non-iteminfo untouched; native failure and empty-decode both fall back
* native decode exposes >100 fields incl. `equip_type_info` / `drop_default_data`, with >3000 equipment records carrying a real (non-`None`) `equip_type_info`
* `match` selects; any-of (#271) unions; expansion emits one fully-resolved `set` per record with the right `_name`/`_key`; non-match intents pass through
* nested: struct fields, list indices, negative indices
* naming: snake鈫抍amel, camel鈫抯nake, `_`-prefixed camelCase, the `_equipAbleHash` fallback, exact-transform-beats-fuzzy, and **ambiguous fuzzy match refused**
* every miss shape returns `None`: bad key, bad root, index past end, scalar-with-path-left, non-integer segment on a list
* dotted real field wins over traversal; `None`-valued field isn't confused with a missing one
* live table: nested match on `drop_default_data.use_socket` selects every socketed item

**227 passed, 0 new failures** across the iteminfo/format3 engine suite.

<sub>One pre-existing failure and one pre-existing stall in the apply-e2e tests on this branch reproduce byte-identically with these commits stashed 鈥?not from this change.</sub>

---

### Update: the new tests weren't actually running (fixed)

Self-audit finding. The three test files this PR adds were all gated on a `tests/fixtures/iteminfo/` directory that **has never existed in this repo**, with a `CDUMM_VANILLA_ITEMINFO_DIR` env-var fallback. So in CI, and on any fresh clone, they skipped silently — and the strongest claims in the PR were guarding nothing:

- 0 opaque records out of 6508
- the enchant-tier oracle (levels must be `0,1,2,...,N-1`) — the only check that can catch a *wrong-but-same-size* layout, which round-trips byte-exact while mis-assigning every field after it
- byte-exact re-serialization of the whole table
- `match` on iteminfo selecting 366 records instead of 0

This is [audit finding C7](https://github.com/faisalkindi/CrimsonDesert-UltimateModsManager/blob/master/tests/fixture_loaders.py) (2026-06-10) recurring, and it takes the same fix that worked then: **commit the table.**

`tests/fixtures/vanilla113/iteminfo.{pabgb,pabgh}.zlib` now sit next to the existing `vanilla110` pair. 5.8 MB → 549 KB, the same ~10:1 those tables get; the already-committed `vanilla110` body is 542 KB, so this adds no new order of magnitude. The 1.10 fixture can't stand in: 1.13 restructured the equipment records (`SubItem` tag 17 + `_enchantDataList`), which is exactly the code these tests exist to pin.

Also turned the layout check into an assert rather than a skip — skipping there would hide a regression in `detect_iteminfo_layout` itself, which is one of the two things the fixture is here to protect.

**Verified with `CDUMM_VANILLA_ITEMINFO_DIR` unset: 33 passed, 0 skipped** (previously 6 of those silently skipped). Full fast suite: 2116 passed, 69 skipped, 0 failed.

> The same fixture + loader also lands in #276, byte-identical. I test-merged both onto `master` in either order: **no conflict** — git resolves identical blobs silently. Each PR stays self-contained and testable on its own.


AgentKush:

AgentKush:

PR #276, opened 2026-07-11
semantic walker: fix the stale ItemInfo schema (iteminfo grid: 11 -> 110 of 113 fields)

AgentKush:
> Independent of #274 / #275 — branches off `master`, touches only the schema-driven walker.

## The bug

The semantic walker **could not walk iteminfo at all**. On the live CD 1.13 table it decoded a median of **11 of 113 fields** before bailing — and it did no better on the committed CD 1.10 fixture. Every consumer degraded accordingly: **the Game Data grid showing ~11 iteminfo columns is this bug.**

## Why nobody noticed

The only test that walks a real iteminfo table looks for its fixture in `tests/fixtures/iteminfo/` — but the fixture is committed at `tests/fixtures/vanilla110/`. So it **always skipped**, in CI and locally. Its docstring still claims *"100% byte-perfect across all 6339 entries"*. That hasn't been true for several game versions.

## What was stale

Every fix is measured against **both** the committed 1.10 fixture and the live 1.13 table, with `iteminfo_native_parser` (which round-trips both byte-exact) as the oracle:

| | was | is |
|---|---|---|
| `ItemIconData` | pre-1.0.4.1 (`u32 + u8 + CArray<u32>`) | fixed **14 bytes** |
| `_materialMatchInfo` | in the field order | **removed in CD 1.10** |
| `_extractAdditionalDropSetInfo` | in the field order | **removed in CD 1.09** |
| two `u8` fields | missing entirely | restored |
| `SubItem` "None" tag | only `14` | also **`16`** (1.10) and **`17`** (1.13) |
| `EnchantData` | 1.11 shape | + the u32 **CD 1.12** added |
| `_enchantDataList` | after `_prefabDataList` | immediately after `_dropDefaultData` |
| `_prefabDataList` / `_gimmickVisualPrefabDataList` | inline | at the **record tail** (1.13) |

`ItemIconData` alone bailed the walk on ~99.9% of records of **both** tables — that's the 11-field ceiling.

The `SubItem` one is worth calling out. Treating tags 16/17 as payload-bearing is what stalled the walk inside `DropDefaultData`. The tell is in the "values" it produced on 1.10 — `65536` (`0x00010000`), `65792` (`0x00010100`), `131072` (`0x00020000`). That isn't data. It's `[socket_valid_count][use_socket][count_lo][count_hi]` read as one little-endian u32 — the socket counters plus the `_enchantDataList` count that follows.

## Result

**Live CD 1.13: median 11/113 → 110/113 fields (10% → 97%).**

## The last 3% is not reachable this way

iteminfo has fields whose presence depends on the **value** of an earlier field — the 12-byte block the native parser injects before `item_desc` on lantern records (`equip_type_info == LANTERN_EQ_TYPE`). The type-descriptor grammar has primitives, CString, CArray, COptional, fixed arrays, substructs and tagged variants; **it cannot express a conditional keyed on a previously-read field.** Closing that last 3% means either extending the grammar or routing iteminfo through the native parser, which already handles it. Not attempted here.

## Scope, stated plainly

The schema is now shaped for **CD 1.13**, the shipping version. It walks CD 1.10 further than before but not fully. That is **not a regression** — it walked neither version before.

I also did **not** repoint the skipping test at the committed fixture. That would turn CI red, since the walker still isn't perfect on 1.10. Silently enabling a test I know fails would be worse than the skip. It's documented instead.

## About the changed test

`test_item_icon_data` is updated, and that deserves scrutiny — "change the test until it passes" is exactly the wrong move. It only ever asserted against bytes it packed itself, so it **restated** the definition rather than checking it, and the definition was wrong. What overrules it is the two shipped tables. New tests pin the sentinels, the icon-data width and the `EnchantData` u32 so this can't silently rot again.

**CI fast job: 2120 passed, 75 skipped, 0 failed.**

---

### Update: the test that should have caught this had a false claim in it (fixed)

`tests/test_iteminfo_walk_real_game.py` opened by claiming:

> `iteminfo: 100% byte-perfect across all 6339 entries`

**Both halves were false**, and had been for at least three game versions. The table has **6508** entries, and the walker completes **zero** of them. It survived because every test in the file was gated on a `tests/fixtures/iteminfo/` directory that has never existed in this repo — so the file skipped in CI and on every fresh clone, and nobody had seen it run since the claim was written.

That's the same rot this PR fixes in the schema, one level up: an unverified claim decaying quietly because the thing that would have caught it was never executing. So the fix is the same — **commit the table** (same `vanilla113` blobs as #275, byte-identical; test-merged both orders onto `master`, no conflict).

#### Measured, not remembered

Now asserted against committed fixtures, so they can't drift again:

| | entries | reaches `_cooltime` | median fields decoded |
|---|---|---|---|
| **1.13** — this PR | 6508 | **6498 (99.8%)** | **110 / 113** |
| 1.13 — `master` | 6508 | 0 (0.0%) | 11 / 113 |
| **1.10** — this PR | 6483 | 0 (0.0%) | 64 / 113 |
| 1.10 — `master` | 6483 | 0 (0.0%) | 11 / 113 |

`master`'s **11 of 113** IS the "iteminfo grid only shows 11 fields" wall.

**On 1.10, this is not a regression and not a fix.** The schema is a 1.13 schema, so on 1.10 the walker runs into `_enchantDataList` — a field 1.13 added and 1.10 doesn't have — and stops there, short of `_cooltime`. It gets *further* than master (64 fields vs 11) and is no worse in target reach (both 0%). `test_1_10_is_a_known_limitation_not_a_silent_one` pins that explicitly, so the limitation is recorded rather than rediscovered later.

**Completing a record is a structural ceiling, not a bug to chase.** The LANTERN block is a 12-byte conditional keyed on the *value* of `equip_type_info`, and the descriptor grammar has no way to express a value-dependent field. Format 3 needs to **reach** the target field, not finish the record — so that's what the tests assert.

#### Two bugs the fixture immediately surfaced

Both had been hidden by the skip:

1. **`test_value_correctness_spot_checks` called `pytest.skip()` if *any* of its tables was missing** — so a missing `vehicleinfo` extract silently discarded the `iteminfo` assertions too. Split in two: the iteminfo half runs off the committed fixture, the vehicleinfo half skips alone.

2. **The end-to-end write proof asserted the *mechanism*** — an 8-byte i64 at a name-end-relative `rel_offset`. On 1.13 that's no longer how the write happens: the writer detects the schema can't decode 1.13 records and falls back to the native relocated-layout writer, which emits a whole-table `_f3_rebuild`. That rebuilt 5.9 MB body is **byte-identical to vanilla except for the 8 bytes of the edit** — i.e. correct, and stricter than the path it replaced — but the old assertion read it as a failure. Rewritten to assert the *outcome* (exactly 8 contiguous bytes change, they are the sentinel, and they land inside record 2200), which holds across both write paths.

The other tables (vehicleinfo / fieldinfo / stageinfo / regioninfo / characterinfo) still need a local extract and still skip — their fixtures run to 25 MB. The docstring now says so, and marks their numbers as **unverified by CI** rather than presenting them as fact.

`walk_real_game`: **11 skipped → 4 passed, 7 skipped.**
Full fast suite, `CDUMM_VANILLA_ITEMINFO_DIR` unset: **2014 passed, 67 skipped, 0 failed.**

---

### Heads-up: this PR had a CI cost, and it's now paid (third commit)

Worth calling out explicitly rather than letting you find it.

Fixing the schema means the walker now decodes **110 of 113 fields instead of 11** — so anything that walks a whole table does roughly **10x the work it used to**. That's inherent to the fix, not a bug in it, but it lands somewhere specific:

`tests/test_f3_whole_table_rebuild.py`, measured per test:

| test | time |
|---|---|
| `test_contaminated_buffer_rebuilds_instead_of_skipping` | 257s |
| `test_header_lets_rebuild_reach_swallowed_records` | 198s |
| `test_clean_buffer_applies_without_rebuild` | 143s |
| `test_name_corruption_refuses_even_with_header` | 142s |
| `test_name_corruption_refuses_rebuild_safely` | 114s |
| **total** | **855s** |

That took the **fast job from 5.4 min → 17m33s**. It passes — but `timeout-minutes: 22`, so it's one slow-runner day from going red on a timeout, on every PR, for reasons unrelated to that PR. Better to fix it now than to have it surface later as a mystery failure.

These are full-table iteminfo round-trips — precisely the category the workflow already routes elsewhere. Its own comment reads: *"~150-530s EACH ... marked `slow` and run in the slow-tests job instead of gating every PR."* At 114–257s each they now sit squarely in that band. They weren't marked before only because the broken walker made a "whole-table" walk cheap **by accident**.

So the third commit adds `pytestmark = pytest.mark.slow` to that module. **Nothing is skipped and no coverage is lost** — the tests move to the slow job (master pushes + nightly + on demand), which is the tradeoff this repo already chose for exactly this class.

Verified:
- `-m "not slow"` collects **0** from that file; `-m slow` collects all **5**
- fast suite: 2014 passed / 384s → **2009 passed / 73s**
- slow suite: **18 passed, 1 skipped, 7m08s** (cap 35 min)
- 0 failures in either

Happy to drop that third commit if you'd rather keep them gating PRs and just raise the timeout — your call, it's a one-line revert.


AgentKush:

PR #277, opened 2026-07-11
feat: gear stats (armour/weapon) are editable on CD 1.13 - they were readable but silently not

AgentKush:
Armour and weapon stats are now editable. They were **readable but silently not editable**, which is the worse of the two failure modes.

> Stacked on #275 (equipment decode + the 1.13 fixture). Review that one first; this PR is the payoff.

## The half that was already fixed

The equipment-record desync fix made stats **readable**: all **3151** equipment records decode, **3129** carry real stat values, **0** opaque.

## The half that wasn't

On CD 1.13 — the only version anyone is playing — the writer takes the "relocated layout" path. **That path only ever did flat field resolution.** The nested-path branch existed solely in the default *pre-1.13* writer.

So every gear-stat intent was counted as an "unwritable field" and dropped, on a record that had decoded perfectly:

```
iteminfo(1.13): 0 of 3 intent(s) applied (0 unknown key, 3 unwritable/opaque field, 0 non-set)
```

The mod applied cleanly, reported success, and changed nothing. No error, no warning, no effect — the same shape of failure as `match` selecting zero records.

**0 of 3 intents before → 3 of 3 after.**

Both writers now go through one shared helper (`apply_nested_intent`). That divergence *is* the bug; keeping two copies of the logic would just reschedule it.

## Path dialect, unified

`match` accepted `list.0.field`; `set` accepted only `list[0].field`. A mod author has no way to guess the two verbs wanted different syntax for the same path. Both now work on both sides.

## Verified against real bytes, not just "it applied"

| check | result |
|---|---|
| 3 edits (both dialects) land their values | ✅ |
| records differing from vanilla | **exactly 1** |
| bytes changed | **exactly 6**, in 3 contiguous runs — one per edit |
| records in → out | 6508 → 6508, **0 opaque** |
| patched table re-serializes byte-exact | ✅ |
| unresolvable path | emits **no change**, never guesses an offset |

(6 bytes because the three values differ only in the low 2 bytes of each i64.)

**Oracle — why the numbers are trustworthy rather than merely well-formed:** enchanting improves an item, so a stat's value across tiers `0..N` must never decrease. **1292 stat series checked, 0 decreasing.** A wrong-but-same-size layout still round-trips byte-exact and still "decodes" — but it scrambles which bytes belong to which tier, and this is the check it cannot survive.

## Stat IDs

15 in vanilla. The game ships **no stat-name table** (they're an exe enum), so most are documented by id only — an invented name is worse than a number.

Three are **evidenced** by the game's own item naming (`Item_Stat_<Set>_<Thing>_LVn` items carry a common baseline pair plus exactly one distinguishing stat):

| id | name | evidence |
|---|---|---|
| `1000007` | CriticalRate | `Item_Stat_AbyssGear_CriticalRate_LV1` |
| `1000010` | AttackSpeedRate | `Item_Stat_AbyssGear_AttackSpeedRate_LV1` |
| `1000011` | MoveSpeedRate | `Item_Stat_AbyssGear_MoveSpeedRate_LV1` |

`1000002` and `1000003` are on nearly everything (9189 / 9377 occurrences) and are almost certainly the primary offence/defence pair — but nothing in the shipped data *proves* which is which, so I haven't named them. Happy to if someone has a source.

I checked the mod ecosystem for a cross-reference and there isn't one: the only stat mod on Nexus for 1.13 (Player Status Modifier) is an **ASI that memory-hooks the runtime stat component** and never touches `iteminfo`. Which is rather the point — nobody mods gear stats through the data tables because until now no tool could decode the records.

`docs/GEAR_STATS.md` documents the paths, the ids, and the fields.

Full fast suite: **2124 passed, 69 skipped, 0 failed.**

---

### Update: the stat names are *in the game* — and the community map is wrong

I said above I wouldn't name `1000002`/`1000003` without proof. Turns out proof exists: **`gamedata/statusinfo.pabgb` maps every stat id to its name.** All 75 of them.

`cdumm.engine.stat_names.load_stat_names()` reads it from the user's install. `STAT_NAMES_CD113` is a snapshot for when the game isn't reachable (CI, fresh clone) — prefer the live read, because a hardcoded table rots, which is the lesson this repo keeps re-learning.

The 15 ids vanilla gear actually uses:

| id | name | | id | name |
|---|---|---|---|---|
| `1000000` | Hp | | `1000012` | ClimbSpeedRate |
| `1000002` | DDD (damage dealt) | | `1000017` | IceResistance |
| `1000003` | DPV (defence) | | `1000026` | Stamina |
| `1000005` | DDV | | `1000027` | Mp |
| `1000007` | **CriticalRate** | | `1000036` | Pressure |
| `1000008` | AttackedDamageRate | | `1000037` | Stamina_UseResourceDecreaseRate |
| `1000010` | AttackSpeedRate | | `1000043` | **GuardPVRate** |
| `1000011` | MoveSpeedRate | | | |

### ⚠️ Don't use the community mapping

`buff_names_community.json` (NattKh/CrimsonDesertCommunityItemMapping) is widely used and is **wrong on 18 of its 24 stat entries** — every one marked `verified: true`. Its resistances are shifted by two:

| key | community | game |
|---|---|---|
| `1000014` | ~~Fire Resistance~~ | Temperature |
| `1000015` | ~~Ice Resistance~~ | Electricity |
| `1000016` | ~~Lightning Resistance~~ | FireResistance |
| `1000017` | ~~HP Regen~~ | IceResistance |
| `1000018` | ~~Stamina Regen~~ | ElectricityResistance |

…and the crit pair is straight **swapped**: `1000006` is CriticalDamage, `1000007` is CriticalRate. **Anyone who set `1000007` intending "critical rate" has been setting critical damage.**

Correction sent upstream: [NattKh/CrimsonDesertCommunityItemMapping#3](https://github.com/NattKh/CrimsonDesertCommunityItemMapping/pull/3) (fixes 18, adds the 51 missing stats).

### Why I believe the game and not the popular file

I went looking for ways to be *wrong*, not ways to be right.

1. **Raw bytes, no parser involved.** In `statusinfo.pabgb` the `u32` key sits physically adjacent to the name string — `key=1000007 strlen=12 name=CriticalRate`. A misaligned decode cannot produce that, which rules out my own parsing as the source of the disagreement.
2. **The game names items after the stat they grant.** `Item_Stat_AbyssGear_CriticalRate_LV1` carries stat `1000007`. Same for AttackSpeedRate (`1000010`) and MoveSpeedRate (`1000011`). Three agreements, zero disagreements.
3. **The community's ids aren't used by any gear.** They put "Guard PV Rate" at `1000019` and "Critical Rate" at `1000006` — **zero** vanilla items use either. The game's GuardPVRate (`1000043`) is on 83 items, **76 of them shields**, and 76 of the 79 equippable shields carry it.

A mapping that names ids nothing uses, while failing to name four ids gear *does* use, is a guess.

Checks 2 and 3 run in CI against the committed fixture, so the snapshot can't quietly go stale.

**What I'm not claiming:** that BlackSpace's internal name honestly describes the in-game *effect*. They could have called the enum `CriticalRate` while it drives crit damage; files can't settle that, only in-game testing can. It doesn't change what you edit, and the docs say so rather than overclaiming.

`stat_label()` renders unknown ids as the bare number on purpose — an invented name reads as authoritative and sends the modder at the wrong stat, which is exactly how the community file did its damage.

Full fast suite: **2131 passed, 69 skipped, 0 failed.**


AgentKush:

AgentKush:

PR #278, opened 2026-07-11
feat(format3): array_append on iteminfo nested lists (sockets, tags, bundles)

AgentKush:
Closes a documented Format 3 gap: `array_append` now works on **iteminfo nested lists**, not just `dropsetinfo.drops`.

> **Stacked PR.** Depends on #274 (equipment decode) → #275 (native match + nested paths) → #277 (1.13 nested-path *writes*). Merge those first; this branch is cut from #277. Its own change is small — `format3_apply.py` + `format3_handler.py` + one test + a docs update. (I've kept all these as `base: master` per the existing pattern; the cumulative diff will shrink to just this once the parents land.)

## The gap

`array_append` was gated to `{("dropsetinfo", "drops")}`. Every other list field got an actionable skip. So to add **one** socket to an item you had to author a `set` carrying the item's *entire* existing list plus your addition — error-prone and verbose.

## The fix

Appending to a nested iteminfo list is mechanically a nested `set` of `current_list + [element]`. iteminfo round-trips byte-exact through the whole-table writer, so this reuses that proven path — grow exactly one record, rebuild the `.pabgh` index, touch nothing else. No bespoke in-place splice.

`_expand_append_intents` runs just before match expansion and translates each iteminfo `array_append` into concrete `set` intents by decoding the record and reading its current list. **It composes with `match`**, so this is one intent:

```json
{ "op": "array_append", "match": { "drop_default_data.use_socket": 1 },
  "field": "drop_default_data.add_socket_material_item_list",
  "new": { "item": 1000080, "value": 5000 } }
```

→ appends a socket-material entry to **every** socketed item (804 records here), each with its *own* current list preserved. A path that doesn't resolve to a list is dropped with a warning, never applied blindly.

## Verified byte-exact on the committed 1.13 fixture

- append by key: exactly one record grows, list == `old + [element]`
- append via `match`: all 804 socketed items, each with its own list + the element
- exactly one record differs from vanilla, **0 opaque, 6508/6508**, `.pabgh` companion emitted
- bare-scalar field refused at validation; `dropsetinfo.drops` and other tables unchanged

## One bug the tests caught

The table-name guard used `_table_name_from_target(target) != "iteminfo"` — but that helper returns the full *path*, not a bare name (the same gotcha the match router hit), so the expansion silently no-op'd on a real target path. Fixed with the same basename normalisation the match decoder uses. The test asserting a real target path *does* expand is what surfaced it.

Full fast suite: **2151 passed, 69 skipped, 0 failed.**


AgentKush:

PR #279, opened 2026-07-11
feat(tools): field-order extractor (macOS reflection strings -> verified _ordered_fields)

AgentKush:
The tooling that turns the macOS binary into verified `_ordered_fields` — the concrete first step toward unlocking the ~82 memory-order-only tables.

> **Stacked on #277** (needs the `schema_verify` harness). Merge that first. Own change is small: one tool, one doc, one test.

## What it does

`tools/extract_field_order.py` scans a Crimson Desert binary for the reflection **error strings** — `<Class>의 _<field>를 읽어들이는데 실패했다` ("Failed to read _<field> of <Class>") — groups them per class **in file order**, and runs the result through `verify_order_source`. It never writes a schema; it **proposes and verifies**.

Method is from NattKh's `PABGB_DECODE_PROCESS.md` (MPL-2.0, cited).

## Proven end-to-end on the real Windows exe

```
reflection classes with field strings: 505
verified tables: 7 | covered: 7 | passed: 0
  ItemInfo   FAIL (ORDER MISMATCH)
  ...
❌ NOT VERIFIED: the extracted order does not reproduce the known tables
   ItemInfo: first divergence at index 0
     known    [0:4] = ['_isBlocked', '_maxStackCount', '_itemName', ...]
     extracted[0:4] = ['_itemName', '_maxStackCount', '_isBlocked', ...]
```

- extracts field **membership** for **505 classes** (essentially every table)
- correctly **warns** it's the Windows binary
- the harness **fails all 7 known tables** — because the Windows string table isn't in read order. **That failure is the tool working**: it refuses to emit an order it can't verify, and shows the first divergence per table.

So: Windows = membership, not order. The **macOS** binary is the one whose strings are ordered (unstripped build — exactly how NattKh decoded `skill.pabgb`).

## The pipeline this creates

```
macOS binary → extractor → {class: [fields]} → verify_order_source
   → must reproduce all 7 known tables → only then add to _ordered_fields
   → still gated to verified_fields after an in-game value spot-check
```

Nothing trusts hand-work (the lesson from the community stat map being wrong on 18/24). Every proposed order proves itself against the tables we already know byte-exact.

`docs/EXTRACT_FIELD_ORDER.md` has the how-to, including fetching just the macOS executable via SteamCMD (`@sSteamCmdForcePlatformType macos`, app `3321460`) without the full install.

`tests/test_extract_field_order.py` pins the string parsing on synthetic input (+4 tests, green); the verification half is covered by `test_schema_verify.py`.


AgentKush:

PR #280, opened 2026-07-11
test: un-skip the tests pinned to one machine — and ship Pillow (Game Data texture preview was dead in the exe)

AgentKush:
Audit finding **C7** — *"a test that can only run on one developer's machine is a test that does not exist"* — was fixed for some tests and **missed for 16**. Those still pointed at `C:\Users\faisa\AppData\Local\Temp\...` and `C:\Users\faisa\Downloads\Compressed\...`, so they skipped in CI, on a fresh clone, and for every contributor — while *looking* like they covered the dropset / skill / iteminfo writers.

> **Stacked on #279.** Merge the iteminfo stack first.

## The reason this matters: it was hiding a shipped-feature regression

**Pillow was never a dependency.** `game_index.decode_image()` imports PIL to decode the game's DDS textures, and `game_data_page.py` calls it — but Pillow was in neither `pyproject.toml` nor the exe. `cdumm.spec` **actively excluded** it:

```python
# PIL/Pillow — not imported by CDUMM (colorthief dep, unused)
'PIL', 'PIL._imaging', ... 'Pillow', 'colorthief',
```

That comment went stale the moment the Game Data preview landed. So `decode_image()` returned `None` and **the texture preview was silently dead in every frozen build**. CI never caught it, because the one test covering it skipped on `No module named PIL`.

Pillow is now a real runtime dependency and is no longer excluded from the build. The DDS decode test runs (36 pass in `test_game_index.py`).

## What changed

Extracted `dropsetinfo` + `skill` from the live game and committed them beside the existing iteminfo tables (~410 KB, same ~10:1 zlib). `fixture_loaders` grows Path-like handles (`vanilla110_file` / `vanilla113_file`) exposing the tiny surface these tests already used (`.exists()` / `.read_bytes()` / `.with_suffix()`), so each test is a **one-line** change — and the skip conditions now evaluate to "present", so the assertions finally execute.

**Layout is now detected, not assumed.** The iteminfo tests called `parse_iteminfo_from_bytes(body)` bare, which uses the module-default field list. That only ever matched one maintainer's 1.11-era extract; against a real table it desyncs — and on a bad count it *spins*. They now call `detect_iteminfo_layout()`, like every non-test caller already does.

## Result: **68 skips → 55, 0 failures**

Newly **running and passing**:
- dropsetinfo apply / format3 / intent / round-trip (8)
- skill list writer (3)
- iteminfo `default_sub_item` trailing block (3)
- the iteminfo **parse + serialize byte-exact trust anchor**, on real CD 1.13 bytes
- the DDS decode test

## What's still skipped — and now says so honestly

- **7 tests** use the vendored **`crimson_rs`** Rust parser, which predates *both* committed fixtures (it fails CD 1.10 and 1.13 identically at offset `0x71`). They now guard on that **real capability** instead of a fake path, so they re-enable themselves if `crimson_rs` is refreshed. Genuine tech debt — now visible instead of disguised.
- **5 tests** are pinned to the *contents* of a pre-1.13 extract (specific record keys; the default-layout `parse_first_record_size` API). **I did not invent new expected values to force them green** — an assertion is only worth something if the number is known-good. The paths they cover are exercised on real 1.13 bytes by `test_iteminfo_gear_stats.py` and `test_format3_array_append_iteminfo.py`.
- **4 tests** need real third-party Nexus mods — other authors' work, deliberately not committed. They now resolve via `tests/fixtures/mods/` or `$CDUMM_MOD_FIXTURES`, so *any* contributor can enable them.

Full suite: **2,174 passed, 0 failed, 55 skipped.**

---

### ✋ Correction — the Pillow claim above is **wrong**, and I want to retract it clearly

I wrote that "the Game Data texture preview was silently dead in every frozen build". **That is not true**, and I should have checked before saying it.

What actually happened: I cut this branch from the iteminfo stack (#274 → #279), which **predates #269**. On *that* branch, Pillow was excluded in `cdumm.spec` and undeclared in `pyproject.toml` — so that's what I saw. But **#269 already fixed both** (`Pillow>=10.3` is declared, and the spec no longer excludes PIL). On `master`, and in the shipped community build, **the texture preview works.**

So there is no shipped regression. I checked the actual `community-3.5.0.2` tag to confirm, rather than leave the claim standing.

**What's left in this PR is still real and still worth merging:**
- 16 test files were pinned to `C:\Users\faisa\...` — one maintainer's machine — so they skipped in CI, on a fresh clone, and for every contributor.
- Committed `dropsetinfo` + `skill` fixtures (~410 KB): **68 skips → 55, and ~15 tests now actually run and pass** (dropset writers, skill list writer, the iteminfo byte-exact parse+serialize trust anchor on real 1.13 bytes).
- Those tests also parsed iteminfo with the *module-default* layout instead of detecting it, which desyncs on any real table. They now detect.
- The remaining skips now state **true** reasons (the vendored `crimson_rs` parser predates both fixtures; 5 tests pin values from an extract that isn't committed) instead of "file not on this PC".

The Pillow edit is dropped from this branch as redundant. Apologies for the noise — better to correct it loudly than let a wrong claim sit in the record.


AgentKush:

PR #260, opened 2026-07-08
fix(iteminfo): apply Format 3 item-price mods (price_list nested path) (#259)

AgentKush:
Closes #259.

`Cheap Gold Bars` and similar price mods target the dotted/indexed path `price_list[0].price.price` (an item's sell/buy value). On CD 1.13 the import failed with:

> field 'price_list[0].price.price' targets a nested struct sub-field (dotted path). Format 3 nested-field writes are not implemented for this field yet.

Two gaps caused it:

1. **Validator gate** — `_diagnose_unsupported_intent` (`format3_handler.py`) has an allowlist of iteminfo nested-path prefixes it accepts (`prefab_data_list[`, `drop_default_data.`, `gimmick_visual_prefab_data_list[`, `docking_child_data.`). `price_list[` wasn't on it, so the intent was rejected at import time.
2. **1.13 writer** — `_build_change_relocated_layout` (`iteminfo_writer.py`, the relocated-layout path used on 1.13) only resolved flat field names, so even past the gate the intent would silently no-op. The full-schema writer already had a nested/indexed path walk; the 1.13 writer didn't.

### Fix
- Add `price_list[` to the iteminfo nested-path allowlist.
- Give the 1.13 relocated-layout writer the same `_resolve_path_target` nested/indexed walk + `shape_matches` gate the full-schema writer uses.

### Verification
- **Byte-exact, live** against the installed CD 1.13 `iteminfo.pabgb`: GoldBar (key 53) `price_list[0].price.price` 50000 → 1 — exactly the 2 bytes of the u64 price change, same table length, **0 collateral records**.
- New CI-runnable tests (`tests/test_iteminfo_price_path.py`, no game fixture) pin the gate acceptance + a still-rejected unknown path + the resolver/assignment/shape gate.
- `272 passed, 32 skipped` across the format3 / iteminfo suite; no regressions.

Scope is intentionally narrow: only `price_list[...]` is added to the allowlist, so the "refuse rather than guess" discipline holds for every other nested path.

Issue #135, opened 2026-05-16
Better Unique Gears failing to apply correctly


mrkillerhomerxD:
When trying to install this mod i get lots of warnings and errors
https://privatebin.net/?2bceb10273e3c9e0#5uGWiei9dC1eN418JSfGnaXAKg4pw8LxBmdokG3jcuas

https://www.nexusmods.com/crimsondesert/mods/2296?tab=posts

faisalkindi:
Thanks mrkillerhomerxD. Couple things to move this forward:

The privatebin link is end to end encrypted in the URL fragment, my side cannot decrypt it programmatically. Could you paste the bundle .txt directly into a reply here as an attachment (GitHub's file uploader takes .txt up to 25MB)? Or, if shorter, copy paste the FIRST 30 or so lines including the TL;DR and the apply log lines that mention warnings.

Better Unique Gears (Nexus 2296) is one of the Format 3 mods I have been investigating this week. BloodGozilla on #121 hit a related bug where Format 3 mods that target iteminfo silently get shadowed when paired with a legacy mod that also targets iteminfo. Fix is on master in commit 4006a8c, ships in v3.3.6. So depending on what warnings your bundle shows, you may already be in the queue for the same fix.

Also still owed you a response on the HAWT mod 237 crash from a couple of weeks back. If you have time, dropping a separate bundle for that one would let me close both threads in one pass.

mrkillerhomerxD:
This sould be the bugreport i saved locally, i thought the pastebin worked :D
[cdumm_bug_report_20260517_011406.txt](https://github.com/user-attachments/files/27898950/cdumm_bug_report_20260517_011406.txt)

i'll check with the HAWT aswell

faisalkindi:
Thanks for the attachment. The bundle tells a clearer story than the error toast did. Two setup issues are stacking and producing the warning noise you saw, but neither one is the apply failing.

[...]

Issue #99, opened 2026-05-11
This MOD ain't workin: Combat God's Plate Gloves Buff


paloroycevincent-sketch:
--- SEVERITY: Bug (wrong behavior) ---

--- USER DESCRIPTION ---
First of all, thank you for the cool mod manager that you made, I find it easy to use and works really well with the game. I appreciate your work and continuous efforts in maintaining and improving this MM. 

1. Dropped zip file mod: Combat God'S Plate Gloves V1.0-2353-1-1778132923
https://www.nexusmods.com/crimsondesert/mods/2353?tab=description
The game starts but this specific mod doesn't work.
Mod won't work, even if its the only mod installed/applied. Tried uninstalling/installing, verify game file integrity>Fix everything>wont work.

3. Presented below info:
<img width="2650" height="425" alt="Image" src="https://github.com/user-attachments/assets/d111e215-1bb5-4018-a514-d96620df9cb6" />
8u 
Apply Completed with Warnings
1 Format 3 mod(s) produced 0 byte changes:
X
Format 3 mod 'Combat God's Plate Gloves' produced 0 byte changes: vanilla bytes for 'iteminfo.pabgb' are unavailable. Most common cause: the live PAZ has been modded by another tool and the vanilla backup is missing.
Revert to vanilla first (Settings -> Fix Everything), then re-apply. If the file is genuinely missing from your install, check the target name spelling or run Steam Verify.
Enabled JSON mods produced no game changes at apply. Run Settings > Fix Everything to rebuild vanilla backups, then re-Apply. If the warning persists, the mods are likely outdated for your current game version.

============================================================
CRIMSON DESERT ULTIMATE MODS MANAGER -- BUG REPORT
============================================================
Generated: 2026-05-11 18:21:06
App Version: 3.3.2

--- TL;DR ---
  No red flags detected in automatic scan.
  OK:
    - Not running as admin (good)
    - All mod-state health checks passed

--- SYSTEM ---
OS: Windows-11-10.0.26200-SP0
Python: 3.14.3 (frozen)
Running as administrator: no
Game Dir: E:\SteamLibrary\steamapps\common\Crimson Desert (exists)
Game Platform: Steam
Game Exe: E:\SteamLibrary\steamapps\common\Crimson Desert\bin64\CrimsonDesert.exe (386.5 MB, mtime 2026-05-03 01:42)
Game Exe Compat Flags: none
Game Version Hash: 473bffe080ef53d7
[...]

PR #281, opened 2026-07-12
fix(format3): validator refused nested iteminfo paths (#259 Cheap Gold Bars — and gear stats)

AgentKush:
Fixes the **Cheap Gold Bars** import failure reported on #259 (Srimk1, lupo1190) — and, in the process, a much bigger one.

## The reported bug

The mod edits `price_list[0].price.price`. Users got:

> field `'price_list[0].price.price'` has no field_schema entry and isn't in the PABGB record schema. **Author needs to add a field_schema/iteminfo.json entry**

— advice the author cannot act on, for a field **the writer has been able to apply for months**. Reported twice, across two releases.

## Root cause: two validator gates that disagreed

There are **two** gates, each with its own hardcoded allowlist of nested-path prefixes:

1. `_diagnose_unsupported_intent` — rejects with "nested writes not implemented"
2. `_classify_intent`'s early-accept — the one that actually returns *supported*

PR #260 added `price_list[` to **gate 1 only**. So the intent passed gate 1, fell through gate 2 to the field_schema lookup, and produced exactly the misleading message above. Two allowlists, grown one bug report at a time (#99, #135, #191, helmysaini, niyaruza, cajae), that had drifted out of sync.

## The part that should sting

**Gear stats were behind the same gate.** `sharpness_data.stat_list[0].change_mb` and the enchant paths were rejected at validation too — so the **entire gear-stat feature was dead end-to-end**, while its tests passed. Those tests called `build_iteminfo_intent_change` directly and never crossed validation.

An allowlist nobody tests is a feature flag nobody knows is off. That's a hole in my own testing as much as in the code, and this reporter is the one who found it.

## The fix

Accept a nested iteminfo path on its **root field existing** on a record (union of field names across every known layout — 118), rather than on an allowlist of full paths. The decoded record is the only thing that knows whether the rest of a path resolves, and the writer already refuses cleanly when it doesn't — 0 byte changes, never corruption.

Deliberately **not** "accept any nested path". That swings to the other extreme: a typo would become a silent *"applied, 0 changes"*, which is worse than the bug being fixed. A real root gets through; a made-up one (`not_a_field.nope`) still errors at import. Both properties are pinned by tests.

`_is_nested_field` is now **one predicate shared by both gates**, matching `iteminfo_writer.is_nested_path` — so they cannot drift apart again.

## Verified on the committed CD 1.13 fixture (validate → write → re-parse)

- price **99,900 → 1** on `CraftingRecipe_Kuku_Golden999K`: 1 record changed, 3 bytes, size unchanged, 0 opaque
- gear-stat set: validates, writes, reads back `7777`
- typo'd root: refused at import
- real root + bogus leaf: validation defers, writer emits nothing

`tests/test_format3_nested_validation.py` goes through `validate_intents` **on purpose** — the step the gear-stat tests skipped.

Full suite: **2,189 passed, 0 failed.**


AgentKush:
Added a follow-up commit to this PR (`5c26786`) — because **this PR is what creates the bug it fixes**, and the two belong in the same reviewable unit.

### What I found

Running the top 59 Nexus mods through the engine (#285) — not a bug report — turned up the mirror image of #259.

This PR is right to kill the hardcoded allowlist: that allowlist was refusing `price_list[0].price.price` and every gear-stat path, and the union-across-layouts rule is the correct replacement. But *"a root that exists in **any** layout CDUMM knows"* includes layouts the player's game **isn't running**. CD 1.13 dropped `prefab_data_list` and `gimmick_visual_prefab_data_list` from the item record, so on the current game:

```
prefab_data_list[0].tribe_gender_list
  validator -> ACCEPTED   ("5098 intents ready to apply")
  writer    -> resolves nothing
  user      -> 0 changes, and no reason given
```

Same two-gates-disagree shape as #259, pointing the other way:

| | validator | writer | result |
|---|---|---|---|
| **#259** | REFUSED | could do it | useless advice |
| **this** | ACCEPTS | cannot do it | **silent no-op** |

An honest refusal at import beats a silent no-op after apply. Right now the second-most-downloaded mod on Nexus (Equip Everything, 13k downloads, *"Requires DMM"*) imports clean against CDUMM and then does nothing.

### The fix

`drop_intents_the_layout_cannot_carry()` in `format3_apply.py`. It has to live in the **apply path** — that's the only place the game's own bytes are in hand (`validate_intents` has no table to look at, and `import_handler` can't cheaply extract vanilla).

It re-checks nested roots against the layout **detected from the installed table**, *not* against the newest layout CDUMM knows. That scoping is the whole point: a player on an older build that really does carry `prefab_data_list` sees **no change in behaviour**. And if the layout can't be resolved it **fails open** — this is a safety net, not a gate.

### Verified on the real game table + real Nexus mods

| intent | before | after |
|---|---|---|
| `prefab_data_list[0].tribe_gender_list` | 2 ready / 0 skipped ← the bug | **0 writable, 2 refused**, field named |
| sockets (#191) / prices (#259) / gear stats (#277) / flat fields | applied | **4 kept, 0 refused** |

No false refusals — every one of those paths was fixed this week and none of them regress.

### What it does *not* do

It does **not** make `prefab_data_list` writable. That needs the field's position in the CD 1.13 record reverse-engineered, which is #285 (the element codecs `_read_PrefabData` / `_write_PrefabData` already exist — only the *position* is unknown). This just makes the failure honest, so the user is told why instead of watching nothing happen.

### Tests

10 new in `tests/test_format3_layout_scoped_roots.py`, including one that **pins this PR's accept-it behaviour** so the fix can't later be mistaken for a no-op, one that pins the fail-open path, and one that pins the `_table_name_from_target`-returns-a-**path**-not-a-name gotcha that already bit #275 (`match` selected zero records) and #278 (`array_append` no-op) — third time it bites, so it's pinned.

Format 3 suite on this branch: **329 passed, 2 skipped**. Full suite on the merged stack: **2221 passed, 55 skipped, 0 failed**. `ruff` adds no new findings.

AgentKush:

Issue #184, opened 2026-05-31
Improvements & Suggestions


devCKVargas:
--- SEVERITY: Visual (UI issue) ---

--- USER DESCRIPTION ---
## 1. Resizable/auto resize Configure window

- At 1440p full screen:
<img width="2560" height="1400" alt="Image" src="https://github.com/user-attachments/assets/358fbdd9-c216-4a52-9e43-05972eab50c8" />
<br><br>

- At 1440p around 50% CDUM window size:
<img width="1587" height="1216" alt="Image" src="https://github.com/user-attachments/assets/bebc5986-d009-4f81-9eb9-78a5f6e1ff3a" />

## 2. Link improvements - Row should be clickable rather than just the text area.

<img width="1255" height="1218" alt="Image" src="https://github.com/user-attachments/assets/f5d02453-a0a3-4ac5-a118-a9bc328c42d9" />

## 3. Double clicking mod should check/uncheck it.

<img width="1169" height="123" alt="Image" src="https://github.com/user-attachments/assets/711a4217-e211-42f1-bea5-2ebfd09b6a27" />

faisalkindi:
Thanks for the screenshots, all three of these are clean asks and I agree with the direction on every one. Tracking them so they make it into a polish pass rather than getting lost in the bug queue.

The Configure window at 1440p is a layout issue I can fix without a redesign, the inner widgets need to be allowed to stretch and the dialog needs a saner minimum size. Doable.

Row click for the mod link cards, that is a UX paper cut I should have caught earlier, the entire row should be the hit target instead of just the text region. Will rewire the click handler.

Double click toggle on mod check state, that one is also a small change to the row delegate, easy to add without breaking the existing click handling.

These will land together in a UX pass rather than spread across separate releases, that keeps the changelog cleaner. No new info needed from you, will ping this thread when the build is up.

faisalkindi:
All three landed on master in commit e92dffc, will ship in the next CDUMM build.

Configure window now scales with the main window. Width floors at max of 500 px or 55 percent of the parent window, height floors at max of 200 px or 45 percent of the parent. Tiny windows still get a usable dialog, big monitors no longer get a sliver against empty space.

About page link cards are clickable on the whole row, cursor switches to a pointing hand so it is obvious you can click anywhere. Open button stays for discoverability.

[...]

PR #282, opened 2026-07-12
fix(ui): make the patch-notes dialog readable (scales with window + spaced-out notes)

AgentKush:
The "What's New" dialog was a small fixed box and the notes read as one continuous blob. Two separate problems, both fixed.

## 1. Size — it never scaled with the window

It was pinned at `minimumWidth 560 / minimumHeight 350` and stayed that size on any monitor. Now it scales with the parent (the same rule devCKVargas asked for on the preset picker, #184):

- floor **760×460** so it's usable on a small window
- **~66% / 58%** of the parent window
- ceiling **1040×760** so lines don't stretch to an unreadable length on an ultrawide

The rule is a pure static method (`_sized_to`) so it's unit-tested without needing a real window.

## 2. Spacing — a wall of text

Each note was one `<li>`: a bold lead-in glued to a long paragraph, `margin-bottom: 3px`, no line-height. Now each note is a **headline line + a detail line**, 150% leading, 20px between notes.

The subtlety that needed a second pass: **Qt's rich-text engine ignores `margin-bottom` on `<li>`.** Put the separation there and the notes end up packed *tighter* than the headline↔body gap within a single note — which reads backwards and still looks like a blob. The gap between notes has to live on the last block *inside* each note. There's a test that pins exactly that (`tail gap > head gap`).

No hardcoded text colour — the `QTextBrowser` stylesheet still owns the theme-aware foreground, so light mode stays readable.

## Verified by eye, not just by asserting on HTML

Rendered the real dialog at 1600×1000 in both themes:

**Dark, latest-only:**
The headline sits on its own line, the detail follows, and notes are clearly separated.

**Light, full history:**
Readable — no washed-out text (an inline `color:` here is what once broke light mode, so there isn't one).

`tests/test_changelog_readability.py` — 14 tests (render split, the spacing hierarchy, version separators, and the sizing floor/scale/ceiling).

Full suite: **2,292 passed, 0 failed.**


AgentKush:

PR #283, opened 2026-07-12
feat(gear-stats): editor that addresses every stat, at its real path

AgentKush:
Armour defence, weapon damage and enhancement values are now editable from the **Game Data** tab. They were already editable by hand-writing a `.field.json` (#277) — this is the UI for it, built on the decode rather than on a guess.

> **Stacked on #281.** Review that first; this PR's own diff is 4 files.

### Why not the editor that already existed

#261/#269 shipped a gear-stat editor whose locator **scanned raw record bytes** for plausible-looking keys (`_STAT_KEY_MIN/MAX`, `MIN_STAT_FREQ = 10`, `_VALUE_ABS_MAX`). Fair enough when equipment records were opaque blobs — there was no other way in. Since #274/#277 there is. Measured against the real CD 1.13 table:

| | |
|---|---|
| Ground truth (native decode) | **28,081** stat entries across 3,319 records |
| The byte scanner | **21,420** — silently drops **6,661 (24%)** |
| Records it disagreed with the truth on | **714 of 3,313** |
| Records it **invented** stats for (they have none) | **33** |

Item **14706** carries `CriticalRate +1` on all 11 enchant tiers; the scanner finds **none** of them — its value floor throws small numbers away. It's been removed from #269 rather than merged, and #261 is closed.

### What this does instead

`gear_stat_view.py` walks a record the parser **already decoded** and returns one entry per stat, each carrying its exact Format 3 path:

```
sharpness_data.stat_list[i].change_mb
enchant_data_list[t].enchant_stat_data.<list>[i].change_mb
```

All **28,081** stats are located, and every one of those paths is asserted to resolve through the writer — the whole table, not a sample.

### Every tier is its own value

The old dialog deduped by stat id, and conceded in its own comment that an edit *"resolves to the **FIRST** located entry with that key"*. A helm carries the same stat id in a dozen places: the base item, and each enhancement tier. So editing "defence" changed the base and left every tier alone — the number drifted the moment the player upgraded the item.

Here each occurrence is its own row (`Base`, `Enhance +3 (flat)`) and its own edit. The dialog says so rather than hiding it. Changing one tier changes one tier; changing all of them is the user's decision, made visibly.

Stat names are read from the game's own `gamedata/statusinfo.pabgb` via `load_stat_names()` — **not** a hardcoded map, because a hardcoded map is exactly what rots when the game patches.

### Tested through the whole chain

`test_editing_two_tiers_of_one_stat_writes_both` goes through `validate_intents` — the gate that silently refused every nested iteminfo path until #281 — and *then* the real writer, and asserts **both** the base and the tier were written.

That's the test the previous gear-stat work never had: its tests called the writer directly and never crossed validation, which is how the feature could be dead end-to-end while its suite stayed green.

- `tests/test_gear_stat_view.py` — 12 tests, against the real 1.13 table
- `tests/test_game_data_gear_button.py` — 4 tests, the page's own plumbing

**Relevant suite: 460 passed, 26 skipped, 0 failed.**


AgentKush:

PR #284, opened 2026-07-12
fix(storeinfo): CD 1.13 moved the stock-record layout; detect it

AgentKush:
Reported by @Srimk1 on #259. donr484's **"Shop Smart. Shop H-Mart"** (Nexus 2626) half-applied — **10 of its 14 stores were silently dropped**:

```
store entry 3101: vanilla stock list does not match the verified layout
(const byte at record offset 34 is 0 (expected 1) at byte 106)
```

> **Standalone.** 6 files, straight off `master`, no dependency on the iteminfo stack. Verified: all 22 open PRs still merge with **0 conflicts** with this one first.

### What was wrong

**CD 1.13 inserted a u32 at stock-record offset 30.** That pushed the flags and the const tripwire down four bytes (head 110 → 114). CDUMM was still reading the **CD 1.11** shape, so the tripwire fired on the *first record of the first store* and the writer refused the batch.

The tripwire did its job — storeinfo has no integrity check, and a misread record written back crashes the game on store open, so refusing was correct. **The bug is that the layout was hardcoded**, and this is the *second* time a patch has moved it (CD 1.11 inserted `is_restore_item`, 109 → 110). Both times every store mod stopped applying until someone hand-edited the constants. Hand-editing constants means the next patch breaks it again.

### The mod author already knew

donr484 had reverse-engineered the field himself. Every record his mod ships carries:

```json
"order_index_113": 4294967295
```

`0xFFFFFFFF` — which is its value in **all 3,661** vanilla records. Two independent sources agreeing is what makes this a fact rather than a guess. Measured on the live table:

| offset | field | |
|---|---|---|
| `@30` u32 | **`order_index_113`** | `0xFFFFFFFF` ×3661 — **new in CD 1.13** |
| `@34–37` u8 | `flag_a` / `flag_b` / `flag_c` / `is_restore_item` | clean `{0,1}` — same as 1.11, shifted +4 |
| `@38` u8 | `const` | `1` (the tripwire) |

The 71-byte opaque value interior **didn't move** — the insert landed before it.

### The fix: detect, don't assume

The layout is now read off the file. Each candidate is trial-parsed and scored by **byte-exact round-trip**, not by "it parsed" — a wrong layout can consume a plausible-looking span and still be misreading it, and byte-exactness is the only test that can't be fooled.

```
CD 1.13   268/293 entries   3661 records   <- picked
CD 1.11   104/293 entries      2 records   <- what CDUMM assumed
CD 1.10     0/293 entries      0 records
```

An older table still detects as its older shape. An unknown shape refuses cleanly. An all-empty store table is *valid*, not unknown (detection scores on entries, not records), so an empty shop isn't a hard error.

### Result — against the real 1.13 game table and the actual mod file

```
pabgb changes emitted : 14 / 14 stores      (Srimk1 got 4 / 14)
patched table re-reads: 268/293 entries byte-exact

Store_Her_General   40 ->  40      Store_Her_Grocery    18 ->  35
Store_Her_Butcher    8 ->  14      Store_Her_Leather   224 ->   7
Store_Her_Tavern    34 ->  23      Store_Her_Equipment  40 ->  20
```

### Tests

`tests/test_storeinfo_layout_cd113.py` — 9 tests pinning the shape, the detection, the byte-exact round-trip of all 3,661 records, the older-build fallback and the unknown-build refusal. Against a **committed 57 KB fixture** of the real table, not an env var.

That matters: the old live-fixture tests read `issue_repro/183/vanilla/`, **a directory that doesn't exist in this repo** — so they skipped silently in CI. That is how this table's layout drifted *two game versions* without anything going red.

And `test_storeinfo_native_parser.py`'s synthetic tests hardcoded `head == 110`. That hardcoding is exactly what let them keep agreeing with a parser that could no longer read the game, so they're now parametrised over every layout.


AgentKush:

PR #287, opened 2026-07-12
feat(iteminfo): decode CD 1.13's relocated prefab_data_list — Equip Everything now applies (2548/2548)

### LINKED ISSUE #285: Gap scan of the top 59 Nexus mods: `prefab_data_list` is not decoded on CD 1.13 (blocks 5,098 intents, incl. Equip Everything 100%)
I ran the **top 59 Crimson Desert mods on Nexus** through the engine against the real CD 1.13 game data, to find gaps before users report them rather than after. Method: pull each mod's newest MAIN file, parse every Format 3 JSON, run `validate_intents` against the live tables extracted from the install (read-only).

**Result: 28 of 32 Format 3 mod files fully supported.** The two fixes from #259 are confirmed on the real mods — Cheap Gold Bars `1/1`, Shop Smart. Shop H-Mart `14/14`.

But it found one big gap, and it's the largest single thing blocking real users right now.

---

## `prefab_data_list` is not decoded on CD 1.13 — 5,098 intents blocked

| Mod | Downloads | Status |
|---|---|---|
| **Equip Everything** (Nexus 2571) | **13,377** | **2,548 of 2,548 intents refused — 100% dead** |
| (1.13) AXIOM-Mask Mega Collection QoL Suite (Nexus 2508) | 22,061 | 2,550 refused |

The refusal is:

```
field 'prefab_data_list' is a variable-length list-of-dicts.
This table doesn't have a list writer yet.
```

**Equip Everything is not a stale mod.** It was updated **2026-07-12**, has 173 endorsements, and its Nexus summary says — in as many words — ***"Requires DMM."*** The author already knows CDUMM can't apply it.

### Why it can't

`iteminfo`'s field list on 1.13 doesn't contain the field at all:

| layout | fields | `prefab_data_list` |
|---|---|---|
| `default` (pre-1.13) | 117 | ✅ present, with `_read_PrefabData` / `_write_PrefabData` |
| `cd113_prefab_relocated` | 115 | ❌ |
| **`cd113_enchant`** ← what the game actually uses | **116** | ❌ (gained `enchant_data_list` instead) |

Measured on the live table: **0 of 6,508 items decode a non-empty `prefab_data_list`**, while the whole table still re-serializes byte-exact — i.e. those bytes are being consumed by the surrounding fields, not read as prefab data.

So the field survived the CD 1.13 patch in the *file*; it just stopped being modelled in CDUMM. That kills two things at once:

* **whole-list sets** — `prefab_data_list` (Equip Everything, AXIOM QoL)
* **nested writes** — `prefab_data_list[N].tribe_gender_list`, which used to work (see the gmVIP233 / niyaruza note in `iteminfo_writer.py`) and is exactly what "Equip Everything" needs for the tribe/gender unlock

### What it would take

The same job as #284, on a bigger table: locate `prefab_data_list` in the 1.13 record, add it back to the `cd113_enchant` layout, and gate acceptance on a byte-exact whole-table round-trip (6,508 records) so a wrong placement can't ship. The read/write element codecs already exist and are unchanged — it's the *position* that's unknown.

I'm happy to take this one unless someone's already on it.

---

## Smaller gap: `inventory.pabgb` has no schema

4 intents in the AXIOM QoL Suite target `inventory.pabgb`, which isn't in `pabgb_complete_schema.json` at all. Low impact, but it's a clean "table not modelled" case rather than a bug.

---

*Everything above is measured against the shipped game, not inferred. The harness is reproducible — happy to contribute it as a CI job so this list is regenerated whenever the game patches, instead of waiting for the bug reports.*

AgentKush:
Closes #285.

**Equip Everything** (13k downloads, *"Requires DMM"*) now applies on CDUMM: **2,548 of 2,548** iteminfo intents, **0 skipped**, and all 2,548 tribe/gender restrictions actually cleared — which is the entire point of that mod.

## What was wrong

CD 1.13 **merged `PrefabData` into `GimmickVisualPrefabData`** and moved the list to the **end** of the item record. CDUMM's 1.13 layout stopped at `repair_data_list`, so the last **76–139 bytes of every record** went unparsed — carried opaquely as `_tail_slack`. The validator then accepted `prefab_data_list` intents, reported them "ready to apply", and wrote nothing.

## Why nothing caught it

> **A byte-exact whole-table round-trip is silent on a missing field.**

All 6,508 records round-tripped byte-exact the whole time — because opaque preservation *is* a perfect round-trip. The gate we lean on hardest proves the bytes are **preserved**, not that they are **understood**. It's still the right gate; it just can't tell you the decoder is *complete*.

So the test here asserts **`_tail_slack` is empty** on all 6,508 records — not merely that the table round-trips. Round-tripping is the check that lied to us.

## How the wire shape was cracked

Brute-forcing grammars got to 96.4% and stalled on 236 records. The **mod itself** broke the deadlock: AerowynX's "Equip All V6" ships *decoded* values for the very struct we couldn't parse. It says `equip_slot_list: [7]` and `[8]` for Folorin_Earring's two elements — and the bytes there read `01000000 0700` / `01000000 0800`.

**`equip_slot_list` is a `u16` array, not `u32`.**

That one fact hid the entire field: an **empty** `u16` array and an **empty** `u32` array are the same four zero bytes. A `u32` reading therefore fits perfectly on the 6,272 records where the list is empty, and desyncs only on the 236 where it isn't. With `u16`: **6,508 / 6,508, zero slack, byte-exact.**

## The writer

Mods only ship `prefab_names` / `equip_slot_list` / `tribe_gender_list` / `is_craft_material`. Post-merge the on-disk element **also** carries `scale`, `animation_path_list` and two flag bytes — and those are **not constant** (44 elements have a non-`(1,1,1)` scale; 68 carry animation entries). Defaulting them would silently rewrite real data, so `normalize_prefab_data_list` merges the mod's element onto the record's own **by index**, letting the mod own only the keys it actually set.

## Verified

| | |
|---|---|
| CD 1.13 | **6,508 / 6,508** byte-exact, **tail slack 0**, 5,687 records expose the field |
| CD 1.10 | still selects the `default` layout — untouched |
| the mod | **2,548 / 2,548** applied; each edited record re-parses with **zero slack** and is byte-stable (write → read → write identical) |

Sockets (#191), item prices (#259) and gear stats (#277) are unaffected — they sit *before* the tail.

## Note for the reviewer

The two whole-table tests are heavy (~35 s each — they read *and* write all 6,508 records). That's deliberate: this is the class of bug that only a full-table, zero-slack assertion catches, and a cheaper test is what let it hide in the first place.


AgentKush:

PR #289, opened 2026-07-12
feat(import): accept .cdmod (crimson-mod-package) — a new Nexus format CDUMM couldn't open at all

### LINKED ISSUE #288: Can't import .cdmod (crimson-mod-package) — a new mod format is live on Nexus
## A new mod package format is appearing on Nexus, and CDUMM can't import it

Found by widening the Nexus sweep after #285 — not from a bug report.

Newly-uploaded mods are shipping **`.cdmod`** files: a format calling itself **`crimson-mod-package`, `format_version: 1`**. Already live:

- [No Fall Damage — .cdmod Rebuild](https://www.nexusmods.com/crimsondesert/mods/3125) (mod 3125)
- [Display Take and Steal Price — .cdmod](https://www.nexusmods.com/crimsondesert/mods/3127) (mod 3127)
- [Crimson Desert .cdmod Converter](https://www.nexusmods.com/crimsondesert/mods/3124) (mod 3124) — a tool whose whole purpose is producing them

## What CDUMM does today

`import_handler.py:278` accepts:

```python
for ext in ('.zip', '.7z', '.rar', '.json', '.bsdiff'):
```

`.cdmod` isn't there. There is no reference to `crimson-mod-package` or `semantic-patch` anywhere in `src/`. So the file can't even be picked up, let alone applied.

## What the format actually is

A **zip** (`PK\x03\x04`) containing:

```
manifest.json
patches/semantic.json
reports/conversion.json
```

`manifest.json`:

```json
{
  "format": "crimson-mod-package",
  "format_version": 1,
  "name": "No Fall Damage - Field JSON",
  "author": "Sportiax",
  "version": "4.4",
  "components": [
    { "type": "semantic-patch", "path": "patches/semantic.json",
      "operation_count": 55, "target_count": 2 }
  ],
  "source": { "format": "format3", "sha256": "36e51f8c..." }
}
```

`patches/semantic.json` — **it's Format 3 under different key names**:

```json
{
  "schema": 1,
  "targets": [{
    "file": "buffinfo.pabgb",
    "operations": [{
      "op": "set",
      "path": "buff_data_list[9].data.variant.body.f01",
      "selector": { "key": 1000185,
                    "string_key": "BuffLevel_Food_FallDamageReduce" },
      "value": 100000,
      "conversion": "conservative"
    }]
  }]
}
```

| `.cdmod` semantic-patch | CDUMM Format 3 |
|---|---|
| `operations[]` | `intents[]` |
| `path` | `field` |
| `selector.key` / `selector.string_key` | `key` / `entry` |
| `value` | `new` |
| `op` | `op` (same) |

Note the manifest even says `"source": {"format": "format3"}` — these are converted *from* Format 3. It's the same model, re-envelloped.

## Why it's worth doing

The mapping is mechanical, so this is small:

1. add `.cdmod` to the accepted extensions (it's already a zip — extraction is free);
2. read `manifest.json`, walk `components[]` of `type: semantic-patch`;
3. translate `operations[] → intents[]` and hand off to the existing Format 3 pipeline.

The interesting part is `path`: `buff_data_list[9].data.variant.body.f01` uses a **`variant`** (tagged-union) segment. CDUMM's nested-path resolver will need to handle that, and it is exactly the generic-variant decoding gap we already know about. So this should be gated on a real `.cdmod` applying end-to-end against the live game table, not just parsing.

## Not doing this blind

Both `.cdmod` mods above are already downloaded and are in the gap-scan corpus, so whoever picks this up has two real files to test against, plus the converter to generate more.

AgentKush:
Closes #288.

A new package format is live on Nexus and CDUMM couldn't open it **at all**. `detect_format()` dispatches on suffix; `.cdmod` matched nothing, fell through to `"unknown"`, and was rejected before extraction.

## What `.cdmod` is

A zip — `manifest.json` + `patches/semantic.json` — whose `semantic.json` is **Format 3 wearing different key names**. Its own manifest says so: `"source": {"format": "format3"}`.

| `.cdmod` | Format 3 |
|---|---|
| `targets[].operations[]` | `targets[].intents[]` |
| `operation.path` | `intent.field` |
| `operation.selector.key` | `intent.key` |
| `operation.selector.string_key` | `intent.entry` |
| `operation.value` | `intent.new` |

So this is a **translation layer, not a second decoder** — it hands off to the existing Format 3 pipeline.

## Why it isn't just mapped to `"zip"`

Because *that* would have been the bug. A plain-zip import extracts it, finds `patches/semantic.json`, doesn't recognise it as Format 3 (no `intents` key), and imports a mod that **changes nothing** — the exact silent-no-op shape of #259, #275, #278 and #285.

So `.cdmod` gets its own branch, and anything that can't be faithfully translated is **refused**, not half-applied:

- unknown package format → refused
- `format_version != 1` → refused (a v2 could reshape `operations`; guessing would silently mis-map fields)
- no `semantic-patch` component → refused, naming what it *did* find
- an operation with no `selector.key` → refused (no record to address — don't invent one)

## Verified against the real files

**`No Fall Damage-4.4.cdmod`**
- → 2 targets, 55 intents; CDUMM's **own** Format 3 parser accepts the translated output
- → against the live CD 1.13 `iteminfo` table: **53 of 53 intents change bytes. 0 silent no-ops. 0 keys missing.**

**`Display take and steal price-1.13.01.cdmod`**
- → correctly **refused**: its only component is a `localization-patch`, not a `semantic-patch`. Fails loudly instead of importing a mod that does nothing. (A second component type worth supporting later — refusing honestly beats pretending.)

*"53 of 53 change bytes"* is the assertion that matters here. Every recent bug in this area sailed past *"N intents ready to apply"* and then wrote nothing.

## Tests

Synthetic package — the real mods are other authors' work and don't belong in the repo. The synthetic one caught a bug the real file **couldn't**: Format 3 *requires* `entry`, and every operation in No Fall Damage happens to carry a `string_key`, so omitting it when absent looked fine against that one file. `entry` is now always emitted.

9 tests, all green. `ruff` clean.

## Not in scope

`localization-patch` components (the second `.cdmod` component type). Refused clearly for now rather than guessed at.

### DIFF
--- a/src/cdumm/engine/cdmod_handler.py
+++ b/src/cdumm/engine/cdmod_handler.py
@@ -1,0 +1,175 @@
+"""`.cdmod` (``crimson-mod-package`` v1) -> Format 3.
+
+A new package format appearing on Nexus (GitHub #288). It is a zip:
+
+    manifest.json
+    patches/semantic.json
+    reports/conversion.json      (optional, informational)
+
+and its ``semantic.json`` is Format 3 wearing different key names. The
+manifest says so itself: ``"source": {"format": "format3"}``.
+
+    .cdmod semantic-patch          Format 3
+    ---------------------------    ------------------
+    targets[].operations[]         targets[].intents[]
+    operation.path                 intent.field
+    operation.selector.key         intent.key
+    operation.selector.string_key  intent.entry
+    operation.value                intent.new
+    operation.op                   intent.op        (same)
+
+So this module translates the envelope and hands off to the existing
+Format 3 pipeline rather than adding a second decoder.
+
+IMPORTANT -- why `.cdmod` must NOT just be mapped to "zip": a plain-zip
+import would extract it, find `patches/semantic.json`, fail to recognise it
+as Format 3 (no `intents`), and import a mod that changes nothing. That is
+the exact silent-no-op shape of #259 / #275 / #278 / #285. Refuse loudly or
+translate properly; never half-accept.
+"""
+from __future__ import annotations
+
+import json
+import logging
+import zipfile
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+CDMOD_SUFFIX = ".cdmod"
+_FORMAT = "crimson-mod-package"
+_SEMANTIC = "semantic-patch"
+
+
+class CdmodError(Exception):
+    """The file claims to be a .cdmod but we can't faithfully translate it.
+
+    Raised rather than returning a partial mod: a .cdmod we only half
+    understand would import clean and silently under-apply.
+    """
+
+
+def is_cdmod(path: Path) -> bool:
+    return path.suffix.lower() == CDMOD_SUFFIX and zipfile.is_zipfile(path)
+
+
+def _read_json(zf: zipfile.ZipFile, name: str):
+    try:
+        return json.loads(zf.read(name).decode("utf-8-sig"))
+    except KeyError:
+        raise CdmodError(f"{name} is missing from the .cdmod package")
+    except (UnicodeDecodeError, json.JSONDecodeError) as e:
+        raise CdmodError(f"{name} is not readable JSON: {e}")
+
+
+def _op_to_intent(op: dict, where: str) -> dict:
+    if not isinstance(op, dict):
+        raise CdmodError(f"{where}: operation is not an object")
+
+    path = op.get("path")
+    if not path:
+        raise CdmodError(f"{where}: operation has no 'path'")
+
+    sel = op.get("selector") or {}
+    if not isinstance(sel, dict):
+        raise CdmodError(f"{where}: 'selector' is not an object")
+
+    key = sel.get("key")
+    if key is None:
+        # Without a key we cannot address a record. Do not guess.
+        raise CdmodError(
+            f"{where}: operation on {path!r} has no selector.key, so there "
+            f"is no record to apply it to")
+
+    # `entry` is REQUIRED by the Format 3 parser, even when empty -- it
+    # raises "intent #0 is missing required key 'entry'" otherwise. Every
+    # operation in the real No Fall Damage package happens to carry a
+    # string_key, so omitting it when absent looked fine against that one
+    # file and only showed up against a synthetic package with none. Emit it
+    # unconditionally; "" is what CDUMM's own mods use when there's no name.
+    return {
+        "key": key,
+        "field": path,
+        "op": op.get("op", "set"),
+        "new": op.get("value"),
+        "entry": sel.get("string_key") or "",
+    }
+
+
+def cdmod_to_format3(path: Path) -> dict:
+    """Translate a .cdmod into a Format 3 (v3.1, multi-target) document.
+
+    Raises CdmodError rather than returning something partial.
+    """
+    if not zipfile.is_zipfile(path):
+        raise CdmodError("not a zip archive")
+
+    with zipfile.ZipFile(path) as zf:
+        manifest = _read_json(zf, "manifest.json")
+        if not isinstance(manifest, dict):
+            raise CdmodError("manifest.json is not an object")
+
+        fmt = manifest.get("format")
+        if fmt != _FORMAT:
+            raise CdmodError(
+                f"unknown package format {fmt!r} (expected {_FORMAT!r})")
+
+        ver = manifest.get("format_version")
+        if ver != 1:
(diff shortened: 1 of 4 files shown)


AgentKush:
## ⚠️ Marking this draft — I found a hole in my own PR

**Do not merge as-is.** The translator is real and tested, but **a user still cannot import a `.cdmod`**, so this PR does not yet do what its title says.

### What I got wrong

I wired `.cdmod` into `detect_format()` (`import_handler.py:416`) and treated that as "the importer now accepts it". It isn't the dispatcher. `detect_format()` is consumed by **exactly one caller** — `mod_compat_check.py:47` — and nothing else.

The actual routing is done by **suffix checks in the GUI**:

```
fluent_window.py:3546   elif p.is_file() and p.suffix.lower() == ".json":
fluent_window.py:3572   elif p.is_file() and p.suffix.lower() in (".zip", ".7z", ".rar"):
fluent_window.py:3659   if p.is_file() and p.suffix.lower() == ".zip":
```

`.cdmod` appears in none of them. So a dropped `.cdmod` is still ignored — the translation code never gets called.

This is the *second* time in this issue I've patched something adjacent to the real path (the first was quoting `prettify_mod_name`'s extension list as the import gate). Same lesson both times: **grep found a plausible-looking line and I didn't check who calls it.**

### What IS done and verified

The engine half stands on its own and is worth keeping:

- `cdmod_handler.cdmod_to_format3()` — translates the package to Format 3.
- Against the real `No Fall Damage-4.4.cdmod`: 2 targets, 55 intents, accepted by CDUMM's **own** Format 3 parser, and **53 of 53 iteminfo intents change bytes on the live CD 1.13 table — 0 silent no-ops**.
- Against the real `Display take and steal price-1.13.01.cdmod`: correctly **refused** (its only component is a `localization-patch`).
- Refuses rather than half-applies on: unknown package format, `format_version != 1`, no semantic-patch component, operation with no `selector.key`.
- 9 tests, green, ruff clean.

### What's left

Route `.cdmod` at the point that actually dispatches: extract → `cdmod_to_format3()` → write a temp Format 3 doc → `import_from_natt_format_3()`. Then the acceptance bar is a `.cdmod` imported **through the GUI** and applied end-to-end — not a unit test of the translator.

Leaving it draft rather than merging something whose title over-claims.

AgentKush:
## Routing wired — and a correction to my own draft note

`db78a51` completes this. **Ready for review.**

### Correcting the comment above

I said `detect_format()` "isn't the dispatcher" and "has exactly one caller (`mod_compat_check.py`)". **That was wrong**, and it was wrong for a dumb reason: my grep used `src/cdumm/**/*.py`, which **doesn't match files sitting directly in `src/cdumm/`** — so it silently skipped `worker_process.py`, the very file that dispatches:

```python
fmt = detect_format(mod_path)
dispatch = {
    "zip":  lambda: import_from_zip(...),
    "7z":   lambda: import_from_7z(...),
    ...
}
handler = dispatch.get(fmt)
```

So `detect_format()` **was** the right place all along. The bug was real — a dropped `.cdmod` genuinely did nothing — but the diagnosis was off: it wasn't "wrong function", it was "**right function, no handler on the other side**". `detect_format()` returned `"cdmod"`, `dispatch.get("cdmod")` returned `None`, and the drop died at *"unsupported file format"* with the translator never called.

That's twice a bad grep has sent me to the wrong conclusion in this issue. The tool was lying in a way I didn't check.

### What landed

- **`import_from_cdmod()`** — translates to Format 3, hands off to `import_from_natt_format_3()`. No parallel apply path.
- **`"cdmod"` entry in `worker_process`'s dispatch table** — the missing half.
- An untranslatable `.cdmod` is **refused at import** (`res.error`), not imported as an empty mod. An empty mod installs clean and changes nothing — the failure shape of #259 / #275 / #278 / #285.

### Tests now pin the routing, not just the function

The previous commit's tests all passed while the feature was unreachable. That's the whole lesson, so it's now guarded directly:

- the dispatch table has a `"cdmod"` entry, and imports `import_from_cdmod`
- the importer matches the dispatch call signature
- an untranslatable package surfaces an error and produces **no** `changed_files`

**12 tests green. No new ruff findings.**

Still verified end-to-end against the real packages: `No Fall Damage-4.4.cdmod` → **53/53 iteminfo intents change bytes, 0 silent no-ops**; `Display take and steal price-1.13.01.cdmod` → correctly refused (`localization-patch`).

AgentKush:

Issue #290, opened 2026-07-13
.cdmod localization-patch components need a .paloc string-table decoder


AgentKush:
## `.cdmod` `localization-patch` components need a `.paloc` decoder

#289 landed `.cdmod` support for `semantic-patch` components and **deliberately refuses** the other component type: `localization-patch`. This is that gap, scoped.

A real Nexus mod is refused today: **[Display take and steal price](https://www.nexusmods.com/crimsondesert/mods/3127)** (`Display take and steal price-1.13.01.cdmod`).

## What the component looks like

`manifest.json`:

```json
{
  "type": "localization-patch",
  "path": "patches/localization/all.json",
  "target": "gamedata/localizationstring_*.paloc",
  "language": "*",
  "change_count": 24
}
```

`patches/localization/all.json`:

```json
{ "changes": [
    { "key": "42597485641824", "op": "append", "suffix": " ({price})" },
    { "key": "42601780609120", "op": "append", "suffix": " ({price})" }
]}
```

So: a u64-hash-keyed string table, and an **append a suffix to the existing string** operation — surgical, rather than replacing the whole table. (That's the point of it: it composes with other language mods instead of clobbering them.)

## Why it's not just wiring

CDUMM has **no `.paloc` parser or writer**. The only `.paloc` awareness anywhere is filename-level: `paz_parse.py:288` rewrites `localizationstring_<from>.paloc` → `localizationstring_<to>.paloc` inside a PAMT node so the game resolves a mod's payload under the right language. The file's *contents* are an opaque blob.

So this needs a real string-table decode/encode:

1. parse `.paloc` — u64 key → string (the `characterinfo_full_parser.py:89` comment already notes "the u64 hash is the key used by the paloc localization table", so the key space is understood);
2. apply `append` (and whatever other ops turn up) per key;
3. re-serialize.

## A useful shortcut, and a caution

The manifest says:

```json
"source": {
  "format": "paloc-full-replacement",
  "vanilla_sha256": "fd1085d8...",
  "modified_sha256": "ece1d343..."
}
```

The converter derived those 24 changes by **diffing a full-replacement `.paloc` against vanilla**. Two consequences:

- **The original mod was a whole-file `.paloc` replacement**, which CDUMM can already install via ordinary file replacement. So users are not fully blocked today — they're blocked from the *surgical, composable* version.
[...]

PR #291, opened 2026-07-13
feat(paloc): decode the .paloc localization table (187,526 records, byte-exact)

AgentKush:
Part of #290.

CDUMM treated `.paloc` as an **opaque blob**. The only awareness anywhere was `paz_parse.rewrite_pamt_localization_filename()`, which renames the file inside a PAMT node for language redirects — the contents were never parsed. So `.cdmod` `localization-patch` components, which *append to individual strings* rather than replacing the whole 24 MB table, couldn't be applied at all.

## The format

```
file   := record*  u32 record_count       ← TRAILER, not a header

record := u32   tag         # category enum
          u32   reserved    # 0 on every record observed
          u32   key_len
          bytes key         # ASCII DECIMAL STRING, e.g. "42597485641824"
          u32   value_len   # BYTES, not characters
          bytes value       # UTF-8
```

Keys being decimal *strings* is why a localization-patch ships `"key": "42597485641824"` as a string and not an int.

## The file self-verifies — and that's the point

The 4-byte trailer is the **record count**. Parse, compare, refuse on mismatch.

This matters more than it looks. **#285 taught us that a byte-exact round-trip proves the bytes are *preserved*, not that they are *understood*** — `iteminfo` round-tripped perfectly for months while carrying 76–139 undecoded bytes per record. Here the format hands us an **independent witness**, so we use it instead of trusting the round-trip alone.

## Do not recompute `tag`

It looks like a length or a checksum. It isn't. Measured across the **187,526** records of the live table:

| hypothesis | matches |
|---|---|
| `tag == value byte length` | 1,049 / 187,526 |
| `tag == value char length` | 2,721 / 187,526 |
| `tag == key length` | 370 / 187,526 |

38 distinct values, max 52 — a category enum. It's preserved verbatim. Had it been value-derived, carrying it through blindly would have **corrupted every string touched**, and the table would still have parsed clean.

Likewise: a change whose key is **absent is reported, not swallowed** (a patch matching nothing would install clean and change nothing — the shape of #259/#275/#278/#285), and an unknown op is **refused rather than guessed**.

## Verified

Against the live **24 MB `localizationstring_pol.paloc`** (187,526 records):

- parses; **trailer agrees**; serialize is **byte-identical**
- append `" ({price})"` to one key → that record changes, `tag`/`reserved` preserved, **all 187,525 others untouched**

**11 tests, green. ruff clean.**

The real-table test runs when `$CDUMM_PALOC` points at one. The file is 24 MB so it isn't committed — the [#290 write-up](https://github.com/faisalkindi/CrimsonDesert-UltimateModsManager/issues/290) shows how to extract one from the Ukrainian Localization mod's PAZ using CDUMM's own `paz` tooling. **No game install needed.**

## Scope — deliberately not wired into `.cdmod` yet

Applying a `localization-patch` means replacing the `.paloc` entry **inside the game's PAZ**, which needs a game install to verify end-to-end. I don't have one, and I'm not shipping an apply path I can't prove.

Until that lands, `.cdmod` packages whose only component is a `localization-patch` keep **failing loudly** (#289) rather than importing as a mod that does nothing. This PR is the codec — the hard part — fully verified against a real table.

### DIFF
--- a/src/cdumm/engine/paloc_handler.py
+++ b/src/cdumm/engine/paloc_handler.py
@@ -1,0 +1,170 @@
+"""`.paloc` localization string tables (GitHub #290).
+
+CDUMM previously treated `.paloc` as an opaque blob -- the only awareness
+anywhere was `paz_parse.rewrite_pamt_localization_filename`, which renames
+the file inside a PAMT node for language redirects. The contents were never
+parsed, so `.cdmod` `localization-patch` components (which append to
+individual strings instead of replacing the whole table) could not be
+applied.
+
+WIRE FORMAT
+-----------
+
+    file   := record*  u32 record_count     <- TRAILER, not a header
+
+    record := u32   tag         # category enum; see below
+              u32   reserved    # 0 on every record observed
+              u32   key_len
+              bytes key         # ASCII DECIMAL STRING, e.g. "42597485641824"
+              u32   value_len   # BYTES, not characters
+              bytes value       # UTF-8
+
+The keys being decimal *strings* is why a localization-patch ships
+``"key": "42597485641824"`` as a string rather than an int.
+
+THE FILE SELF-VERIFIES
+----------------------
+The 4-byte trailer is the record count. Parse and compare: if they disagree,
+the framing is wrong. That check is free, and it is the thing that stops this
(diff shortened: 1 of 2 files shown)


AgentKush:

Issue #292, opened 2026-07-13
Conflict list flags mods as conflicting when they edit different items in the same table (false positive)


AgentKush:
Reported by **falobos76** on #191 against community build 3.5.0.5.

> the conflict list lists ALL of them as conflicting, thus only the first one in activation order (the highest one) wins, which in my case is "Armor Five Sockets". I checked in-game, but it seems that all of them apply their modifications to the various items and weapons.

He's right, and he's spotted something important: **the apply is correct, the conflict report is wrong.** All of pinapana's socket mods really do apply together in game. CDUMM just tells the user they don't.

That's the worst kind of false positive — it pushes people to disable mods that were working.

## Root cause

`conflict_detector.py:251` — when two mods touch the **same entry** in a file, a `Conflict` is raised unconditionally:

```python
if a_entries and b_entries:
    shared_entries = a_entries & b_entries
    if shared_entries:
        for entry_path in sorted(shared_entries):
            ...
            conflicts.append(Conflict(... winner_id=winner_id ...))
    else:
        # Different entries in the same PAZ — compatible
        conflicts.append(Conflict(..., level="paz",
            explanation="...but different game files inside it. Compatible."))
```

The only "compatible" verdict is for mods touching **different files**. Two mods that both edit `iteminfo.pabgb` are always declared to conflict, with a winner — even when one sets `use_socket` on helmets and the other sets it on gloves.

The granularity is **file/entry**. It needs to be **record + field**.

## Why the apply path disagrees

Format 3 mods don't overwrite a table; they carry per-record, per-field intents, and `apply_engine` merges them. Two mods editing different item keys (or different fields on the same key) compose perfectly — which is exactly what falobos76 observed in game.

So CDUMM already *knows* the intents. The conflict detector just isn't looking at them.

## The fix

For Format 3 mods, compare on `(target_file, record key, field)`:

- **different keys** → compatible, no conflict
- **same key, different fields** → compatible, no conflict
- **same key, same field, different values** → a real conflict; report the winner
- fall back to the current entry-level comparison for byte-offset (Format 2) mods, where CDUMM genuinely can't tell what a patch touches

## Acceptance bar

Not "the list looks shorter". Use his actual mod set:

1. pinapana's socket mods (which edit different item ranges) enabled together → **zero** conflicts reported, and all of them still apply;
2. two mods that genuinely set the **same field on the same item key** → still reported, with the correct winner.

[...]

PR #294, opened 2026-07-13
fix(apply): refuse byte-offset mods on a table a Format 3 mod rebuilds (game-corrupting)

### LINKED ISSUE #293: Game crashes when a byte-offset mod and a Format 3 mod both target iteminfo (stale offsets after table rebuild)
Reported by **falobos76** on #191 against community build 3.5.0.5.

> Regarding the crash on startup, I verified what happens if the pinapana patches and one of the following mods are activated simultaneously:
> 1) No Cooldown Infinite Durability Stamina Spirit (Nexus 2714)
> 2) True Thief Gloves - No Cooldown (hand-edited for 1.13.01)
> 3) mission_efficiency_x20.json (hand-edited for 1.13.01)
>
> ... I use all the mods on DMM, and they all work.

Each mod works **alone**. The game crashes only when a pinapana mod is combined with one of these. That combination is the clue.

## Hypothesis (NOT yet proven — do not fix on this alone)

The two groups are different *kinds* of mod:

- **pinapana's socket mods** are **Format 3**. CDUMM applies them by parsing `iteminfo.pabgb`, editing records, and **re-serializing the whole table**. Records change size, so **every byte offset after the first edited record moves**.
- The three mods above are **Format 2 byte-offset patches** (`{game_file, changes:[{offset, original, patched}]}`) — including two that falobos76 hand-edited to *fix the offsets for 1.13.01*.

If a byte-offset patch for `iteminfo` is applied against a table that a Format 3 mod has already rebuilt, its offsets point at the wrong bytes. Writing there produces a structurally invalid table — which is exactly the sort of thing the game would crash on at startup.

This would explain all of it: each works alone; DMM works because it applies both kinds semantically rather than by raw offset.

## What to check first (before writing any fix)

1. Does `apply_engine` apply Format 2 offset changes and a Format 3 whole-table rebuild to the **same** `game_file` in one pass? (`apply_engine.py` ~L445-532 tags Format 3 whole-table targets in `f3_target_files`.)
2. Is there a guard that re-verifies a Format 2 patch's `original` bytes at its `offset` **after** a rebuild? There is a signature check (`apply_engine.py:500`), but confirm it fires in this order — if it does, the user should be getting a clear refusal, **not a crash**.
3. Reproduce with his exact set. His mod list is attached to #191 ([MOD.txt](https://github.com/user-attachments/files/29945818/MOD.txt)).

## If confirmed, the right behaviour

CDUMM must **refuse, loudly**, not corrupt:

> "*Mission Efficiency x20* patches `iteminfo.pabgb` at fixed byte offsets, but *Armor Five Sockets* rebuilds that table, so those offsets no longer point where the mod expects. Applying both would corrupt the file. Disable one, or ask the author for a Format 3 / .cdmod version."

Silently writing to stale offsets and letting the user find out when the game won't start is the worst possible outcome — and it's the same family as #259 / #275 / #278 / #285: **CDUMM reported success and the result was wrong.** Here it's worse than a no-op; it's corruption.

The proper long-term fix is to re-anchor offset patches against the rebuilt table (find the record the offset fell inside, re-locate it), but **the refusal must land first** — a user should never be able to reach a crashing game state this way.

AgentKush:
Closes #293. Reported by **falobos76** on #191.

**This one corrupts.** It doesn't merely fail to apply — it hands the user a game that won't launch. So it goes ahead of everything else in the queue.

## What happens

A **Format 3** mod doesn't patch bytes. CDUMM parses the whole table, edits records, and **re-serializes it**. Records change size, so **every byte offset after the first edited record moves**.

A **Format 2** mod patches *fixed offsets*. Applied against a table a Format 3 mod has rebuilt, its write lands in the middle of some other record. The table is structurally invalid and the game will not start.

falobos76 hit exactly this: pinapana's socket mods (Format 3, `iteminfo`) **plus** any of three byte-offset mods that also patch `iteminfo` → crash on startup. **Each works alone.** He also notes they all work in DMM — which fits: DMM applies both kinds semantically, so it has no raw offsets left to go stale.

## Why nothing stopped it

CDUMM **already knew** which tables get rebuilt. `f3_target_files` is collected in `aggregate_json_mods_into_synthetic_patches` (L445–454) and then used for **exactly one thing** — a label in the display log (L532). Nothing guarded on it.

The nearest protection compares *signatures between two Format 2 mods* and only emits a `logger.warning`; it never sees the Format 3 side at all.

## The fix

Pre-scan the enabled mods for the tables any Format 3 mod rebuilds, then **refuse** to aggregate a byte-offset patch against one of them — naming **both** mods, and carrying the refusal out in `_refused_offset_mods` so the apply path and the bug report can surface it.

Refusing loses one mod's changes. **Not** refusing loses the user's game.

This is the same family as #259 / #275 / #278 / #285 — *CDUMM reports success and the result is wrong* — except worse. Those were silent no-ops. This is silent **corruption**.

## The trap it would have fallen into

Format 3 ships `iteminfo.pabgb`. Format 2 ships `gamedata/iteminfo.pabgb`. Compared raw they **never match** — and *a guard that never matches is a guard that isn't there*. That exact path-vs-bare-name mismatch made `match` select zero records (#275) and `array_append` no-op (#278). Third time, so `_normalize_target()` and its test are pinned.

## Not refused (a guard that over-fires is its own bug)

- an offset mod on a **different** table → still applies
- offset mods with **no** Format 3 mod enabled → untouched; this is the case that has always worked and must keep working

**8 tests, green.** The `E402`/`F401` ruff findings in `apply_engine` are pre-existing (L229–232), not from this change.

## Scope

Does **not** fix the false-positive conflict list (#292) — that's separate, and reporting-only. And the proper long-term answer here is to **re-anchor** offset patches into the rebuilt table (find the record an offset fell inside, re-map it). But the refusal has to land first: nobody should be able to reach a crashing game state while that's being built.

### DIFF
--- a/src/cdumm/engine/apply_engine.py
+++ b/src/cdumm/engine/apply_engine.py
@@ -360,6 +360,17 @@ def _rewrite_mount_error_with_mod_names(
     )
 
 
+def _normalize_target(name: str) -> str:
+    """Compare a Format 3 `target` with a Format 2 `game_file` fairly.
+
+    One ships ``gamedata/binary__/client/bin/iteminfo.pabgb`` and the other
+    ``gamedata/iteminfo.pabgb``. Comparing them raw silently never matches --
+    the exact path-vs-bare-name trap that made `match` select zero records
+    (#275) and array_append no-op (#278). Compare on the bare table name.
+    """
+    return (str(name or "").replace("\\", "/").rsplit("/", 1)[-1].lower())
+
+
 def aggregate_json_mods_into_synthetic_patches(
     db, overlay_priority_tiebreak: bool = True,
 ) -> tuple[dict, list[dict]]:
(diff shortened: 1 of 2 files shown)


AgentKush:

PR #296, opened 2026-07-13
fix(apply): re-anchor byte offsets onto a Format 3 rebuild — 231/231 of falobos76's real changes now land correctly

AgentKush:
The **proper** fix for #293. Builds on #294 (which made the corrupting combination *refuse*); this makes it **work**.

## Tested with falobos76's actual mods

Not fixtures. His socket mod rebuilding the **live CD 1.13 `iteminfo` table**, plus **mod 2714 `03 - Infinite Durability Only.json`** — the exact variant he uses:

```
live CD 1.13 iteminfo : 5,938,891 bytes, 6,508 records
Format 3 rebuild      : 5,938,891 -> 5,939,051 (+160 bytes)
mod 2714 (his file)   : 231 byte-offset changes, all 231 valid against vanilla

WITHOUT the fix : 229 of 231 write to the WRONG bytes   <-- his corruption
WITH the fix    : 231 of 231 land on the author's exact bytes, 0 refused
```

**229 of his 231 changes were landing in the wrong place.** That's the crash, reproduced from his files.

## How it works — no schema needed

The Format 3 whole-table change already carries both halves:

```
{"offset": 0, "original": <entire vanilla body>, "patched": <entire rebuilt body>}
```

and a record the Format 3 mod **didn't touch is byte-identical in both**. So there's no need for the record index or any per-table knowledge: locate the patch by the bytes *around* it.

Four things keep it honest:

1. the patch's `original` must be present at the old offset **in vanilla** — otherwise the mod was already broken, built for another game version;
2. context is taken **before** the patch only — a window running past it would break on a *later* edit and falsely refuse a patch that never moved;
3. displacement is **bounded** by the table's total size delta;
4. the `original` bytes must be present at the **new** offset — a remap that doesn't land on the author's bytes is not a remap.

## The bound is the interesting part — and it came from my own bug

My first version refused **76 of his 231 changes** as *"ambiguous"*. Item records are repetitive, so a context window legitimately matches in several places, and I refused rather than guessed. Correct — but useless: he'd have lost a third of his mod.

Records only shift because *earlier* records changed size, so **|new − old| can never exceed the table's total size delta** — **160 bytes** here, while the spurious matches are *megabytes* away. Bounding the search kills the ambiguity without weakening anything: we still require **exactly one** candidate inside the bound, so we're never choosing between two plausible answers. A unique anchor found *outside* the bound is refused too — a patch can't have moved further than the table can have shifted, however unique that match looks.

Result: **76 → 0** refusals, with no loss of safety.

## Still refused (correctly)

A record whose **bytes** the Format 3 mod changed. The two mods genuinely disagree about them, and silently picking one is the same bug class as #259 / #275 / #278 / #285.

**Not** refused: a record that merely **grew** while the patched bytes stayed put. A guard that over-fires is its own bug, and there's a test for that direction too.

## Tests

12, green, ruff clean. Deliberately synthetic — **if the algorithm needed the real iteminfo layout to be testable, it would be the wrong algorithm.** The real-table run above is the end-to-end proof.

## Relationship to #294

#294 stops the corruption; this makes the mods actually work together. #294 should still land first — it's the seatbelt, and it's simpler to review.

### DIFF
--- a/src/cdumm/engine/offset_reanchor.py
+++ b/src/cdumm/engine/offset_reanchor.py
@@ -6,10 +6,16 @@ instead of corrupting the game. This makes it WORK.
 THE PROBLEM
 
 A Format 3 mod doesn't patch bytes: CDUMM parses the whole table, edits
-records and re-serializes it. Records change size, so every byte offset
-after the first edited record moves. A Format 2 mod's fixed offsets then
-point into the middle of some other record -- the table is invalid and the
-game won't start.
+records and re-serializes it. Records change size, so every byte offset after
+the first edited record moves. A Format 2 mod's fixed offsets then point into
+the middle of some other record -- the table is invalid and the game won't
+start.
+
+Measured on falobos76's actual files (his socket mods + mod 2714 "Infinite
+Durability Only", against the live CD 1.13 iteminfo table):
+
+    231 of his byte-offset changes anchor correctly in vanilla
+    229 of those 231 would write to the WRONG bytes after the rebuild
 
 THE FIX, WITHOUT NEEDING TO UNDERSTAND THE TABLE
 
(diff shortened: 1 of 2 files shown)


AgentKush:

PR #297, opened 2026-07-13
feat(convert): v2 byte-offset mod -> Format 3 field-name mod (822/822 on real mods, verified byte-for-byte)

AgentKush:
> **Stacked on #287** (CD 1.13 iteminfo decode + the committed 1.13 fixture). The last commit is the only new one; review that. Without #287 the tests would *silently skip* on master, which is the exact failure this project keeps finding in its own suite.

Asked for by **falobos76** on #191. Two of the mods he depends on are **abandoned on Nexus**, so every Crimson Desert patch moves the bytes out from under them and someone has to re-measure the offsets by hand — or the mod is simply dead.

## The idea

| | |
|---|---|
| **v2 change** | "at byte 4,192,880, replace `64 00` with `FF FF`" |
| **Format 3 intent** | "set `max_endurance` to 65535 on item 1002862" |

The second one **never goes stale.** CDUMM looks item 1002862 up by key in the player's own files on every Apply, so Pearl Abyss can move it wherever they like. Convert a mod once, and it survives every future game patch — including for mods whose authors are long gone.

## Measured on his real mods, not fixtures

Against the live 1.13 item table (5,938,891 bytes, 6,508 records):

```
03 - Infinite Durability Only            231/231 converted, verified byte-for-byte
02 - 1 Second Cooldown Only              180/180 converted, verified byte-for-byte
10 - CD + Durability + Stamina + Spirit  411/411 converted, verified byte-for-byte
```

**822 of 822.** And the field names come out right without anything telling the tool what the mods are called — "Infinite Durability" resolves to `max_endurance` + `is_destroy_when_broken`; the cooldown variant adds `cooltime`. That correspondence is the whole trick working.

## Safety, in order

1. **Every change's `original` must match vanilla at that offset**, or the mod is stale for this game version and the **whole** conversion is refused. Stale offsets don't fail loudly — they name the *wrong item*, confidently.
2. **Partial-field writes.** The cooldown floats patch two bytes of four. The patched bytes are overlaid onto the field's current bytes and the *whole field* is decoded, so the intent carries the resulting **value**, not a byte fragment. (A converter that skipped this would emit a different number and pass a naive review.)
3. **A change that rewrites a record's key is refused.** A Format 3 intent addresses items *by* key, so `set key` can never apply. Found by a test of mine that aimed at the wrong field: the writer refused it — the right answer — but the converter should say so in words rather than emit an intent guaranteed to do nothing.
4. **Nothing is silently dropped.** Anything it can't name is reported with a reason, and any unconverted change **blocks verification**.
5. **It checks itself.** The converted mod is applied and required to reproduce the original v2 mod **byte-for-byte**. If it doesn't match, no file is written. Nobody has to take the conversion on trust.

The output is a real `.field.json` on disk — the user keeps it, re-imports it, backs it up, or sends it to the mod author.

## What this deliberately does NOT do

**Recover a mod that is already stale.** I tested it rather than assuming, and the answer is genuinely interesting:

- **Whole-field writes** → the reconstruction is *forced*. On the durability fields, matching on values alone reproduces the mod exactly, and the single extra record it finds turns out to be an item **added to the game after the mod was written** (idx 6470, sitting immediately after the mod's last item).
- **Partial-byte writes** → it over-matches. On the cooldown floats it pulls in **6 records out of 66 that the mod never touched**, sitting *in the middle* of its range — so they're not new items, they're false positives. Six items would silently get 1-second cooldowns.

That's a guess wearing a lab coat, and it's the same bug class as #259 / #275 / #278 / #285 — each of which was a plausible-looking write to the wrong place. So: **convert a mod while it still works**, and the question never arises.

## Tests

9, against the **committed CD 1.13 fixture** (6,508 real records), not a hand-made table. They pin the refusals as hard as the conversions — a stale mod, an un-nameable field, and an ID rewrite each have a test, because the refusals are the part that protects someone's save.


AgentKush:

PR #298, opened 2026-07-13
fix(conflicts): the same table is not the same edit — stop flagging mods that work together (#292)

### LINKED ISSUE #292: Conflict list flags mods as conflicting when they edit different items in the same table (false positive)

AgentKush:
Closes #292. Reported by **falobos76** on #191:

> the conflict list lists ALL of them as conflicting, thus only the first one in activation order (the highest one) wins ... I checked in-game, but it seems that all of them apply their modifications to the various items

He's right — and he spotted the important half: **the apply is correct, the report is wrong.** All of pinapana's socket mods really do apply together in game. CDUMM just told him they didn't.

That's the worst kind of false positive: it pushes people to disable mods that were working.

## Root cause — deeper than the issue said

I filed #292 saying the granularity was file/entry and needed to be record+field. Reading the code, it's worse:

**A Format 3 mod stores no delta and no byte ranges.** Its `mod_deltas` row is a stub whose `byte_start` / `byte_end` are the **PAZ entry's** — identical for every mod targeting that table. So the detector had nothing to compare *but* the filename, and two mods on `iteminfo.pabgb` were **always** declared to conflict, with a winner.

And `_semantic_conflict_info` — the function meant to give the richer field-level verdict — reads a delta metadata key, `semantic_fields_changed`, that **nothing in this codebase has ever written**. It always returned `None`. The "semantic" verdict has never once fired since the day it was added.

The intents were there the whole time, in `mods.json_source`. The detector just wasn't looking.

## The fix

Compare on `(target, record key, field)`:

| | verdict |
|---|---|
| different items | **compatible** — no winner, status stays clean |
| same items, different fields | **compatible** — apply merges per field (pinapana ships *All EQ Dyeable* explicitly to be used *with* a socket mod) |
| same item, **same field** | **real conflict** — named, and resolved by load order |

## Two more bugs this shook out

Both would have made the fix worse than the bug:

1. **The `semantic` level was handled by neither status function.** `get_mod_status` and `get_all_mod_statuses` check for `byte_range` and `paz`; anything else falls through to `clean`. Since `semantic` was dead code it never mattered — but *emitting* it without fixing this would have reported a genuine field-level conflict as **CLEAN**. One false verdict traded for another.

2. **A `match` intent parses with `key=0`, not `None`.**
   ```
   Format3Intent(entry='', key=0, field='price', op='match', new=1, old=None)
   ```
   My first guard keyed "records we can't know" off `key is None` — so a `match` mod would be read as *"edits item 0"* and cheerfully declared compatible with everything. Caught by the test that exists precisely to pin that fallback. Now gated on the **op**, not the key.

## What still gets the conservative verdict

`match`, `clone_record`, `new_record`, `delete_record` — the record set isn't knowable without the game files — and any non-Format-3 mod, which has no intents to read. Those keep the old behaviour exactly.

**We narrow the warning only where we can prove it. We never invent a compatibility we can't demonstrate** — that's the same discipline as #294/#296, and the reason this project's warnings are worth reading at all.

## Tests

8, pinning the refusals as hard as the compatibility: a real same-field conflict must still fire, must never read as clean, and an unknowable `match` mod must fall back rather than be waved through. Plus the path-vs-bare-name trap (`gamedata/iteminfo.pabgb` vs `iteminfo.pabgb`) that already caused #275 and #278.

Existing conflict tests: 31 passed, no regressions.

### DIFF
--- a/src/cdumm/engine/conflict_detector.py
+++ b/src/cdumm/engine/conflict_detector.py
@@ -13,6 +13,12 @@ from cdumm.storage.database import Database
 
 logger = logging.getLogger(__name__)
 
+#: Format 3 ops that name the record they edit, so we can say up front
+#: which items a mod touches. `match` selects by predicate; clone_record /
+#: new_record / delete_record add or remove records. For those, the record
+#: set isn't knowable without the game files — so we don't claim to know.
+_KEYED_OPS = frozenset({"set", "array_append"})
+
 
 @dataclass
 class Conflict:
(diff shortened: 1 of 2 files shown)


AgentKush:

Issue #190, opened 2026-06-03
Character Creator Female rapier and shield module import issue


woowoots:
The new module : Female Rapier and Shield Module.json cant be imported in CDUMM for now : 

There is the report : 

```
============================================================
DIAGNOSTIC REPORT: Female Rapier and Shield Module.json
============================================================
Import error: Format 3 mod targeting equipslotinfo.pabgb: none of the 2 intent(s) can be applied yet.

=== equipslotinfo.pabgb ===
2 intent(s) skipped:
  - 2x target 'equipslotinfo.pabgb' has no schema in CDUMM (table 'equipslotinfo' not in pabgb_complete_schema.json, and field not in field_schema/equipslotinfo.json)

Workaround: use the offset-based JSON variant of this mod if one exists. That format works in CDUMM today.

Type: File (.json)
Size: 0.0 MB
```

i've copied the report, not sure if there is any others files i should add here.

thanks

lurkser:
The zipped mod in general can't be imported since there's no select window for us to pick the race option in CDUMM

faisalkindi:
I get what you mean about the race picker not showing. To actually fix it I need the mod, because the picker decides whether to appear from how the mod packs its variants, and I cannot tell what is different about this one from the description alone. Could one of you attach the zip here? If it is too large, paste the folder and file layout inside it, just the structure and the folder and file names. That shows me why the variant detector is missing the race options so I can get the picker to appear.

woowoots:
https://www.nexusmods.com/crimsondesert/mods/837?tab=files&file_id=10646

there is the mod direct download link from nexus, its going to be faster :)

faisalkindi:
The race and gender picker is fixed in v3.3.18, which is out now. I pulled Character Creator from Nexus and traced it. The mod ships the few JSON option files right next to the per-race body folders (Human, Goblin, Orc, male and female), and CDUMM was jumping straight to the JSON options and never showing the body folders, so you never got to choose. Now the race and gender picker comes first. Grab v3.3.18 and re-drop the zip and you should see it.

Separate from that, the Female Rapier and Shield Module.json from the first post fails for a different reason. It edits a game table called equipslotinfo that CDUMM does not have a layout for yet, so it skips those changes instead of applying them. That part is not in this build. I have the file and I am mapping that table by hand, which is slow going, and I will post here when it is ready.

faisalkindi:
[...]

PR #304, opened 2026-07-19
Table coverage + DMM Mod Builder parity: statusinfo, inventory, characterinfo, match-all/$in, stale-cache resilience

AgentKush:
Seven more self-contained fixes from the community fork, cherry-picked cleanly onto `master` (v3.6.0). Companion to #303 (which carried the import / macOS / notification batch). Community version/changelog bumps are excluded so nothing downgrades v3.6.0. Net: 6 source files + 7 test files + 2 fixtures (+1292/-38). All from the #191 DMM-mod thread.

## New table writers
**statusinfo `stat_level_data` — DIRECT SPEED-class stat mods**
statusinfo has two record shapes (71 regular @84B tail; 4 rate @212B tail carrying `stat_level_data` = 16×i64 at tail offset 80). Length-preserving element write that refuses non-rate records. Byte-exact.

**inventory `.pabgb` slot counts — DMM "max inventory" mods**
Applies slot-count writes to `inventory.pabgb`.

**characterinfo `call_mercenary_cool_time` — no-dragon-cooldown mods**
Applies the mercenary-call cooldown field.

## DMM Mod Builder parity
- **match-all `{}`** selects all records; **`$in`** (`{key: {$in: [...]}}`) selects any-of.
- **scale-intent tolerance** — a DMM `scale` intent that lacks `new` no longer fails the whole mod; that one intent is skipped and the rest apply.

## 1.14.00 stale-cache resilience
The 1.14.00 update didn't change table formats — only the vanilla cache went stale. Two fixes: the converter reads vanilla via the live game index when the cache is stale, and a stale index is detected by offset (not only by a decode error). Fixes the `.pabgb` crash and `.pabgh` silent-garbage.

## Verification
- 64 targeted tests green on the v3.6.0 base; new statusinfo/inventory fixtures included.
- Table writes are byte-exact round-trip verified.
- Full suite runs in CI on this PR.

*Not included* (still blocked, tracked separately): `op=scale` multiply and `[*]` wildcard nested paths — gimmickinfo has no decoder and dropsetinfo's raw fields don't map yet.

Refs #191 (statusinfo DIRECT SPEED, inventory slots, characterinfo cooldown, and the DMM match-all/$in/stale-cache fixes are all from that thread).


AgentKush:
**Merge-order note.** This PR is independent of #303 (no shared files). It **does** overlap #305 on exactly one file — `src/cdumm/engine/characterinfo_writer.py` (`_FIELD_MAP`) — so whichever of #304/#305 merges *second* will conflict there.

Trivial resolution: keep #305's refactored 4-tuple map + `_NEW_SCHEMA_MAP`/discriminator, and re-add this PR's `call_mercenary_cool_time` entry in the 4-tuple form:

```python
"call_mercenary_cool_time": ("_callMercenaryCoolTime_offset", 0, "<Q", 8),
```

Order otherwise doesn't matter; #303 merges cleanly with both.

AgentKush:
> **Stacked on #304** (which introduces `inventory_writer`). Base it there, or merge #304 first.

## The bug

**AXIOM's "Inventory Space 240/700" (Rayruii) applies zero bytes.**

Not a missing decoder. `inventory_writer` has framed this table correctly since the DMM "max inventory" work — it just wasn't reachable. It was registered only under the **legacy** names `default_slot_count` / `max_slot_count`, and the Mod Builder later dropped the `_count` suffix. So every current-vintage inventory mod writes `default_slots` / `max_slots` and misses the writer entirely:

```
inventory.pabgb: supported=0 skipped=2
  SKIP default_slots: target 'inventory.pabgb' has no schema in CDUMM
  SKIP max_slots:     target 'inventory.pabgb' has no schema in CDUMM
```

Same rename drift that hit characterinfo.

## The fix

Three aliases in `_SLOT_FIELD_OFFSET` plus their `LIST_WRITERS` entries. **Additive** — the legacy names stay registered, so mods already in the wild are untouched, and a test pins that both spellings emit byte-identical changes.

## Verified against an independent oracle

Not our own inference. **donr484's *Max Inventory Storage* (Nexus 1561) ships the same edits as raw byte-offset patches with hand-written labels, built for CD1.15.00.** All 11 of its labelled offsets reproduce byte-exact against the shipped table, and its labels name the vanilla values.

Running AXIOM's mod through the **real** apply path (`expand_format3_into_aggregated`) against a live 1.15.00 install now writes:

```
offset=3193 span=3193..3198 (6B)
   +0 abs=3193 same       0 ->    0
   +2 abs=3195 CHANGED   50 ->  240   <- donr484 "backpack 50->700"
   +4 abs=3197 CHANGED  240 ->  700   <- donr484 "backpack max 240->700"
   original bytes match the live table: YES

changed u16s on oracle-proven offsets: 2   elsewhere: 0
```

Both offsets are his, derived independently of anything in this repo. Vanilla reads `(50, 240)` exactly as his labels claim.

## What this nearly cost — worth reading

I first derived a **head-anchored** model for the pair: `u32 count + count×3 items + 8-byte gap + <u16 default><u16 max>`. It fitted four entries cleanly (`PearlCharacter` (20,240), `Quest` (300,300), `InvisibleInventory` (5,5), `PearlUser` (20,240)) and looked solid.

It was wrong. The real positions run from **+106 to +2788** — nowhere near the head. Those four matches were coincidence, and a writer built on them would have written into unrelated fields on the record every mod actually targets.

Going and finding another author's labelled offsets is what caught it. Two independent derivations then agreed on the same anchor (`28 80 02 00 00` with the slot triple immediately before it), which is why this is shippable and the earlier version wasn't. **"It explains several entries" is not evidence.**

## Tests

`tests/test_inventory_current_dmm_names.py` — 7 tests:

- current names registered; **legacy names still registered** (no regression for shipped mods)
- validator accepts the current names
- vanilla `Character` reads (50, 240) — donr484's labelled values
- the AXIOM mod writes **exactly** the oracle offsets and nothing else
- aliases and legacy names produce byte-identical changes
- every oracle offset sits at the marker-relative position the model claims

Also commits the vanilla extract (`inventory.pabgb`/`.pabgh`, **3.9 KB compressed**) so these run in CI instead of skipping on a fresh clone. The test loads it with `zlib` directly rather than via `tests/fixture_loaders`, so this PR carries no dependency on #310.

15 passed with `test_inventory_writer`. Ruff clean on the new test and on `inventory_writer`; `format3_handler`'s one pre-existing `F821` is unchanged before and after.

## How it was found

A category-wide sweep of the Nexus corpus through the real apply path. Of 103 Format 3 mod-files judged across all 6 categories that ship Format 3 mods, 86 applied cleanly and the dead ones collapsed to 3 tables — this one, plus `equipslotinfo` (#310) and `buffinfo` (#311).

AgentKush:
## Second gap in this PR: skill indexed element paths

Timuela's **"Focus Aerial Roll doesn't cost Spirit"** applied **zero bytes** in every variant it ships. Two separate causes, both fixed here.

### 1. Only the legacy whole-list names were registered

`LIST_WRITERS` carried `_useResourceStatList`. Current DMM addresses **one element and one field inside it, by letter**: `use_resource_stat_list[0].d`. Now registered under the wildcard key `use_resource_stat_list[].d` — the same `[N]` → `[]` normalization #190 added for equipslotinfo.

### 2. The apply side matched the raw field, not the wildcard

```python
list_routable = [i for i in intents if (table_name, i.field) in LIST_WRITERS]
```

`validate_intents._routable` normalizes, `format3_apply` didn't. So even once registered the intents **passed validation and then produced no change** — validate-then-do-nothing, the same failure mode as #310 and #311. This would have bitten any future indexed registration too, so the fix is a shared `_has_list_writer` helper rather than a skill-specific patch.

### The letters come from the parser, not from me

`_parse_resource_stat` in the vendored parser spells out the element — 22 bytes, matching the `cnt * 22` stride the record walker already uses:

```
a stat_type u8 | b stat_hash u32 | c flag  u8
d value     i64| e hash2     u32 | f hash3 u32
```

So `.d` is `value`, the resource cost — vanilla **−20000** on all three dash skills the mod targets.

### This also settled a disagreement I couldn't have resolved by inspection

The mod's own two variants write **different values** for the same field:

- `.field.json` → `1`
- DMM V3 export → `18446744073709551615`

The field is `<q` (i64), so the second is **−1**, not a colossal cost. DMM emits signed values as unsigned. Written raw it wouldn't even fit `<q` and the whole-table write would fail, so the writer reinterprets via two's complement.

### Verified

Through the real apply path against a live 1.15.00 install: all three variants go from 0 bytes to a real patch, and the whole-table change differs in **exactly 24 bytes = 3 skills × one 8-byte i64**. The writer refuses any element index or leaf the record doesn't actually have, rather than creating a field the game never had.

`tests/test_skill_indexed_element.py` — 16 tests covering the letter↔wire-order mapping, the signed conversion (including `18446744073709551615 → -1`), registry/parser anti-drift, and refusal of malformed paths and of `buff_level_list` (a list *of lists*, which has no letter mapping yet and must not be guessed).

47 passed / 4 skipped across the skill, inventory, equipslotinfo and format3-apply suites. Ruff clean on both touched files including `--select I`.

---

**Status of the three dead tables** found by the category sweep: `inventory` ✅ (above), `skill` ✅ (this), `buffinfo` — still can't apply, but #311 makes it say so honestly instead of blaming the user's game version. That one needs the variant body decoded and is genuinely its own job.

faisalkindi:
Reviewed this in an isolated worktree. **Holding the merge on one defect — the statusinfo writer. Everything else in the PR is sound and I want it in.**

## statusinfo: `_SLD_TAIL_OFFSET = 80` is off by four, and the element type is wrong

I decoded the PR's own committed fixture (`tests/fixtures/vanilla113/statusinfo.pabgb.zlib`) both ways. There are 4 rate records with a 212-byte tail.

Read as 16 x int64 at tail offset 80, three of the four decode to garbage:

```
AttackSpeedRate  @80 int64[0:4] = [-4611686018427274623, -9223372036854560965, ...]
CriticalRate     @80 int64[0:4] = [-4611686018427274623, -9223372036854560965, ...]
DHIT             @80 int64[0:4] = [-6917529027641081849,  8070450532247928847, ...]
```

Read as 16 x u64 32.32 fixed-point (value = u64 / 2**32) at tail offset **76**, all four come out clean:

```
DHIT          : 7.8125, 15.625, 23.4375, 31.25, 39.0625, 46.875 ...
                deltas: 7.8125 x9, then 3.90625 x5   <- exact constants
MoveSpeedRate : 78125.0, 156250.0, 234375.0, 312500.0 ...
                deltas: 78125.0 x5, then a smooth decay
AttackSpeedRate / CriticalRate: clean monotonic ramp
```

Three things corroborate 76 independently:

1. Element [15] decodes to exactly **0.0 in all four records**.
2. `tail[204:212]` is **byte-identical across all four records** (`41dc11f323878d49`) — plainly a trailer, not per-record array data.
3. `76 + 16*8 = 204` partitions the 212-byte tail exactly. The offset-80 reading runs to 208 and overlaps that trailer.

The current docstring actually contains the tell: it notes a clean non-decreasing ramp *in the low 32 bits*. The conclusion to draw from that is that the element boundary sits 4 bytes earlier, not that the high bits are junk.

## Why the tests don't catch it

MoveSpeedRate is the one record that still looks plausible at offset 80, because its values are small enough that the 4-byte shift cancels when the high dword is zero. And `tests/test_statusinfo_writer.py:85-89` derives its containment bound from the same `+ 80` constant, so it structurally cannot detect the overrun.

## Concrete consequence

Running the writer with the real DIRECT MOVEMENT SPEED payload (all 16 elements set):

```
tail[204:208]   vanilla: 41dc11f3   ->   modded: 00000000
```

The `stat_level_data[15]` write lands at `tail[200:208]` and zeroes four bytes of a field that isn't part of the array and is never zero in vanilla. That fires on exactly the presets this PR targets.

There's a second, sharper failure too: `_pack_i64` accepts the full 0..2^64 range, but the high dword lands in the next element's fraction dword. Setting only `stat_level_data[0] = 10000000000` silently yields `1410065408.0` and reports success with `dropped == []` — a wrong value written and reported as fine.

## What I'd like

`_SLD_TAIL_OFFSET = 76`, elements treated as 32.32 fixed-point (or at minimum a refusal for values >= 2^32 rather than silent truncation), and the tests re-pinned against the decoded ramp instead of raw low-dword integers.

I deliberately haven't patched this myself: what a DMM mod means when it sets a plain integer here is a semantics question about the mods, not just the bytes, and guessing at it is how the wrong constant got locked in the first place. Your call on the encoding.

## The rest of the PR is good, and verified

- **inventory** — marker framing independently checked on both fixtures (1.13 and 1.15): exactly 20 named records, exactly 20 markers, every record's first marker inside its own bounds. Decoded values are real (`Character` 50/240, matching the known 240-slot overflow limit; `Wagon` 1/12; `CampWareHouse` 240/1000). Properly evidenced. One note: `inventory_writer.py:114`'s `find` has no upper bound, so a future table shipping a record without the marker would walk into the next record and patch its slot counts — worth bounding to the next record start and refusing.
- **characterinfo `call_mercenary_cool_time`** — clean, with a genuine byte-exact round-trip.
- **Reachability** — no dead code. `build_statusinfo_changes`, `build_inventory_changes` and `parse_indexed_element_field` all reached from `apply_engine.py:798`; I probed `validate_intents` directly and it returns supported=1 for each new field rather than relying on the tests.
- **No silent no-ops** — both writers return `(changes, dropped)` and warn on empty or dropped.
- Tests I ran: statusinfo/inventory/characterinfo 32 passed; DMM match + stale-cache set 55 passed; broad regression 491 passed, 9 skipped, 0 failed.

Split the statusinfo writer out and the remainder can go in immediately, or fix the constant here — whichever you prefer.

One more thing you'll want to know: I merged #305 ahead of this, so `characterinfo_writer.py` now conflicts. #305 refactored `_FIELD_MAP` to 4-tuples `(offset_key, byte_delta, struct_fmt, width)` where the write offset is `record[offset_key] + delta`. Your `call_mercenary_cool_time` entry becomes `("_callMercenaryCoolTime_offset", 0, "<Q", 8)` — delta 0 is correct and I verified it three ways (the parser sets that offset immediately before its own `<Q` read at the same position, and a measured run emitted change offset 64 against parser offset 64). Put it in `_FIELD_MAP`, not `_NEW_SCHEMA_MAP`, so legacy mods keep it — `SUPPORTED_FIELDS` derives from `_FIELD_MAP` and will pick it up automatically.

AgentKush:
I shall fix this when I get home from work today 😊

AgentKush:
Thank you for this — decoding the fixture both ways in a worktree rather than reading the diff is what caught it, and you were right to hold the merge. `80` is wrong and the tests couldn't have found it. Taking the split you offered: **I'll pull statusinfo out so the rest can merge now.**

I went to verify `76` before changing the constant, and it doesn't hold either. **The array starts at 68.**

## Confirming your findings first

All three of your corroborations reproduce exactly:

- element `[15]` at 76 decodes to `0.0` in all four rate records
- `tail[204:212]` is byte-identical across all four — `41dc11f323878d49`
- `76 + 16*8 = 204` partitions the tail exactly, and the offset-80 read runs to 208, into that trailer

And your fixed-point call is right — the values are unmistakable once read as 32.32:

```
DHIT          @76: 7.8125, 15.625, 23.4375, 31.25, 39.0625, 46.875 ...
MoveSpeedRate @76: 78125.0, 156250.0, 234375.0, 312500.0 ...
```

## Why 68 and not 76

`68 + 17*8 = 204` also partitions the tail exactly, so the boundary argument doesn't separate them. What does separate them is the head of the array:

```
MoveSpeedRate  @68: 0.0, 78125.0, 156250.0, 234375.0 ...
DHIT           @68: 0.0,   7.8125,   15.625,  23.4375 ...
```

A leading `0.0` and a trailing `0.0`, with 15 real values between — a level-0 slot and a terminator. At 76 that leading zero is outside the array and the first element is already a live value.

The mods settle it. `2xAtkSpd_2xMovSpd` writes 16 values:

```
MoveSpeedRate  mod: 0, 40000000, 80000000, 120000000, 160000000, 200000000 ...
vanilla @68       : 0.0,  78125.0,  156250.0,  234375.0,   312500.0,  390625.0 ...
vanilla @76       : 78125.0, 156250.0, 234375.0, 312500.0, 390625.0, 468750.0 ...
```

The mod's index 0 is `0`, and vanilla's slot 0 is `0.0` — **at 68**. At 76 the mod's `0` would blank a live value, and the mod's `stat_level_data[15]` (its maximum, `500000000`) would land on the `0.0` terminator. The mod's array shape mirrors vanilla's at 68 and contradicts it at 76.

Same at every index: mod `[j]` tracks vanilla `[j]` at 68 across all four records, off by one at 76.

## The part I'm not going to guess

Your instinct to leave the encoding to me was the right one, and having looked at it I don't think I can answer it from bytes either.

Under 32.32, vanilla `MoveSpeedRate[1]` is `78125.0` and the mod writes `40000000` there. Read as a fixed-point *value* that's 512x vanilla; read as a raw u64 it's `0.0093`, about 1/8400 of vanilla. Neither is "2x move speed", which is what the mod is called and what its users report getting. So there is a unit conversion between what the Mod Builder writes and what the field holds that I can't derive from the table alone — and picking one anyway is exactly how `80` got locked in.

So: **statusinfo comes out of this PR.** I'd rather ship nothing for those presets than a third guess at the constant. It goes back to a branch of its own, and it doesn't come back until I can check a written value in-game rather than argue about it from a ramp.

Two notes for when it does:

- `_pack_i64` accepting the full u64 range is a real bug independent of the offset — you're right that `stat_level_data[0] = 10000000000` yielding `1410065408.0` with `dropped == []` is a wrong value reported as success. Whatever the encoding turns out to be, that becomes a refusal.
- The test bound deriving from the same `+ 80` constant is the structural version of the same mistake I made on inventory: a check computed from the thing it's checking. The replacement pins against decoded values and the trailer, so a wrong offset fails rather than moves the goalposts.

## The rest

- **inventory `find` bound** — agreed, fixing it. Unbounded `find` walking into the next record is the same class as the `WareHouse`/`CampWareHouse` substring problem the envelope match already guards; the marker search deserves the same treatment. It'll be bounded to the record and refuse rather than reach past it.
- **characterinfo** — taking your 4-tuple form into `_FIELD_MAP`, `("_callMercenaryCoolTime_offset", 0, "<Q", 8)`. Thanks for verifying delta 0 three ways and for pointing out `SUPPORTED_FIELDS` derives from `_FIELD_MAP` — that would have been an easy thing to get subtly wrong by putting it in `_NEW_SCHEMA_MAP`.
- Noted on #305 merging first; I'll rebase rather than merge so the history stays readable.

Genuinely useful review — the offset-80 defect was mine and the fixture I committed is what let you prove it, which is the system working. I'd rather find out this way than from a user whose save had a zeroed trailer.

AgentKush:
Done — pushed as `15cc0e8`. PR is mergeable again.

**statusinfo is out.** Writer, tests, fixtures, the `_WHOLE_TABLE_TARGETS` entry, the apply branch and the `_STATUSINFO_SLD_RE` early-accept all removed. It comes back on its own branch once a written value can be checked in-game rather than argued from a ramp.

**characterinfo** — rebased onto v3.7.0 with your 4-tuple form in `_FIELD_MAP`: `("_callMercenaryCoolTime_offset", 0, "<Q", 8)`.

**inventory `find` bound — I owe you an honest no on this one.** I implemented it and it broke 6 of 24 inventory tests. Scanning forward for the next `<u32 len><name><0x00>` envelope lands *inside* the current record: records run to ~2.8 KB and their interior bytes contain runs matching that shape, so the bound falls before the real marker and all 20 working records get refused. A correct bound needs the record boundaries derived some other way. I've left the limit and that failed approach documented at the call site rather than ship a bound that breaks what currently works — your call whether that's acceptable for this PR or whether you'd rather it block.

Fast suite after the split: **2498 passed, 42 skipped**.

---

One more thing on statusinfo, for whoever picks it up. I went looking for a labelled oracle and found that mod 2511 publishes, for the **same** game version (1.13.0), both the JSON V3 field mod *and* a "Mod Folder Version" PAZ overlay — i.e. the already-modded `statusinfo.pabgb` next to the intents that produced it. Diffing those two answers both the offset and the encoding outright, no inference.

I got as far as pulling both files and locating `gamedata/statusinfo.pabgb` in the overlay's PAMT, but `_extract_from_paz` hands back ciphertext for it (8185 of 8216 bytes differ and the values are noise), so the overlay is encrypted and the diff needs that path sorted first. Recording it here because it turns the encoding question from a judgement call into a measurement, which is what it should have been from the start.

AgentKush:
Following up on the encoding you left to me — **I stopped guessing and measured it.** You were right that it was a semantics question about the mods; it turned out to be answerable, just not from the table alone.

## The oracle

Mod 2511 publishes, for the **same game version (1.13.0)**, both the JSON V3 field mod *and* a "Mod Folder Version" PAZ overlay containing the already-modded `statusinfo.pabgb`. So the intents and their result ship side by side, and the diff answers offset and encoding outright.

The overlay is ChaCha20-encrypted — and **CDUMM already decrypts it**. `paz_crypto.decrypt(raw, "statusinfo.pabgb")` opens it fine. It doesn't happen automatically because `PazEntry.encrypted` is a filename heuristic covering `.xml/.css/.html/.js`, so a mod-authored `.pabgb` overlay reads back as noise (8185 of 8216 bytes "differ") and looks like a corrupt file rather than ciphertext. Worth a separate fix — `_encrypted_override` exists for exactly this case.

Decrypted, the overlay's head matches vanilla exactly and **109 bytes differ**. Those are the mod's edits.

## The answer

```
_SLD_TAIL_OFFSET = 68          (not 80, not 76)
16 elements of u64
raw_u64 = mod_value * 2**24    (not 2**32, and not raw)
```

**144 of 144 elements match `raw == asked * 2**24` exactly**, across all six presets the mod ships (2x/3x/4x AtkSpd, each alone and paired with 2xMovSpd). Zero mismatches.

Vanilla decodes to clean round numbers under the rule, which is the corroboration:

```
AttackSpeedRate : 0, 29000000, 55000000, 78000000, 99000000, 118000000 ...
MoveSpeedRate   : 0, 20000000, 40000000, 60000000, 80000000, 100000000 ...
DHIT            : 0,     2000,     4000,     6000,     8000,     10000 ...
```

And the presets are literally what their names say: **2xAtkSpd sets index 1 to 58000000 against vanilla 29000000 — exactly double.** That's the "2x", and it's what closes the semantics question rather than arguing it.

## On 76

Your three corroborations for 76 all reproduce, and they're all real — `element[15]` is `0.0`, `tail[204:212]` is a byte-identical trailer, `76 + 16*8 = 204` partitions exactly. `68 + 17*8 = 204` partitions exactly too, so the boundary test couldn't separate them; what separates them is the leading `0.0` slot and, decisively, the overlay.

The 32.32 reading was the near-miss: it produces a clean ramp because 2²⁴ and 2³² differ by a constant factor, so *any* consistent scale looks tidy. The overlay is what pins the exponent.

## Next

statusinfo comes back on its own branch with offset 68, `value * 2**24` encoding, a refusal when the product overflows u64 or the record isn't a 212-byte rate record, and tests pinned against decoded values rather than raw low dwords. The `_pack_i64` full-range bug you found is fixed as part of it — `stat_level_data[0] = 10000000000` silently yielding `1410065408.0` with `dropped == []` becomes a refusal.

Thanks for holding the merge and for not patching the constant yourself. If you had, we'd have shipped 76 and it would have been wrong in a way the tests couldn't see — same shape as the 80 defect, one step over.

AgentKush:
statusinfo is back as #319, on offset **68** with a **24-bit fixed point** encoding — settled by measurement, not by a third guess at the constant.

Your 76 reproduces every corroboration you cited, and they're all real; what they can't do is separate 76 from 68, because both partition the 212-byte tail exactly and any consistent scale yields a tidy ramp (2²⁴ and 2³² differ by a constant factor). Mod 2511 ships the JSON field mod *and* a PAZ overlay of the resulting table for the same game version, so the request and its result sit side by side. At 68 all 144 written elements satisfy \
aw == asked << 24\; at 76, none do. Running the mod's own intents through the full pipeline reproduces the author's shipped table byte for byte on all six presets.

The 212 bytes now fully account for themselves: 68-byte header, array at 68..195, zero terminator at 196, then the 8-byte constant at 204 — the thing 80 was overrunning into, exactly as you found.

Your \_pack_i64\ finding is fixed there too: 10000000000 round-trips whole instead of coming back as 1410065408 with nothing reported dropped.

Reading that overlay needed one unrelated extraction fix, up separately as #318 — no dependency between them.

Issue #302, opened 2026-07-18
[Mod]: Character creator female animation v3


woowoots:
### Mod name + Nexus page

https://www.nexusmods.com/crimsondesert/mods/837

### Mod version

7.6

### Mod type (if known)

JSON / Format 3 patch

### Skipped-patch / error text

```text
Import Complete
Format 3 mod imported: 9 intent(s) ready to apply, 16 can't yet:

=== characterinfo.pabgb ===
9 intent(s) ready to apply
16 intent(s) skipped:
- 4x field 'appearance_name' has no field_schema entry and isn't in the PABGB record schema. Author needs to add a field_schema/characterinfo.json entry mapping 'appearance_name' to a tid or rel_offset
- 4x field 'character_prefab_path' has no field_schema entry and isn't in the PABGB record schema. Author needs to add a field_schema/characterinfo.json entry mapping 'character_prefab_path' to a tid or rel_offset
- 4x field 'F35' has no field_schema entry and isn't in the PABGB record schema. Author needs to add a field_schema/characterinfo.json entry mapping 'F35' to a tid or rel_offset
- 3x field 'default_action_action_index' has no field_schema entry and isn't in the PABGB record schema. Author needs to add a field_schema/characterinfo.json entry mapping 'default_action_action_index' to a tid or rel_offset
- 1x field 'character_weight' has no field_schema entry and isn't in the PABGB record schema. Author needs to add a field_schema/characterinfo.json entry mapping 'character_weight' to a tid or rel_offset
```

### Your game version

1.14.00

### Does the mod work in another manager (e.g. DMM) on the same game version?

Haven't tried

### CDUMM bug report

```text

```

woowoots:
The new json format with the intend thing to patch json value dont seem to currently work with the latest CDUMM version (3.6.0)
went back of using the preview female animation version with the offset patching to have this being applied and function

lurkser:
Also when importing the entire zip, character creator 7.6 doesn't show the pop-up to select the race and gender anymore.
It still worked when I manually extracted the HumanFemale folder from the archive though, but the female animation mod still didn't work.

AgentKush:
Thanks @woowoots and @lurkser — there are two separate things here, and both are on me to fix:

[...]

Issue #150, opened 2026-05-20
Format 3 characterinfo.pabgb - nested field writes and missing schema entries not supported


shaurya1188:
**Bug Report: Format 3 mod import fails for characterinfo.pabgb**

**Mod:** Female Animations.json (Kliff PHW animations by AerowynX/Cracker, v1.1)

**Issue:** All 15 intents are skipped on import with the following errors:

1. **Nested struct writes not implemented**
   - Fields `upper_chart.group_lookup` and `lower_chart.group_lookup` use dotted paths targeting nested struct sub-fields. Format 3 nested-field writes are not supported for these yet.
   - Affects 6 intents (3x each field)

2. **Missing field_schema entries**
   - Fields `skeleton_name`, `lookup_25`, and `flag_c` have no entry in `field_schema/characterinfo.json` and are not recognized in the PABGB record schema.
   - Affects 9 intents (3x each field)

**Expected behavior:** Intents should apply successfully, or the loader should provide a clear path to resolve schema gaps.

**Suggested fixes (from the error output):**
- Implement nested struct sub-field writes for dotted paths in Format 3
- Add `skeleton_name`, `lookup_25`, and `flag_c` mappings to `field_schema/characterinfo.json`

**Workaround used:** None available — no offset-based variant of this mod exists currently.

**Loader version / CDUMM version:** v.3.310

Please add this version or format to supported formats 

[Female Animations.json](https://github.com/user-attachments/files/28046645/Female.Animations.json)

faisalkindi:
Fixed in v3.3.11, just released. CDUMM can now write the five characterinfo fields the mod uses (upper_chart.group_lookup, lower_chart.group_lookup, skeleton_name, lookup_25, flag_c). I verified the field positions against the vanilla game file rather than guessing, and tested that all 15 intents apply. Update CDUMM, re-import Female Animations, and it should apply. Reopen if not.

faisalkindi:
No reply within 24 hours after the retest request. Closing on the assumption v3.3.11 (and now v3.3.12) resolved it. If the fix does not work for you on the current build, reopen with a fresh bug report bundle from Settings then Bug Report and I will dig in.

PR #305, opened 2026-07-19
characterinfo: apply current-DMM appearance/prefab fields (Character Creator 7.6) (#302)

AgentKush:
Character Creator / Female Animations 7.6 (Nexus 837) patches `characterinfo.pabgb` with the **current** DMM Mod Builder's semantic field names — `appearance_name`, `character_prefab_path`, `skeleton_name`, `lookup_24`, `lookup_25`. DMM renamed the action-chart slots between versions, so those resolve to **different** block offsets than the legacy #150/#192 naming. The result on 3.6.0: the 9 "ready" intents were written to the **wrong** offsets and the mod didn't take (the author reverted to raw offset-patching), and the appearance/prefab intents were skipped outright.

### Pinned against the real table, not guessed
The mod copies the **Damian** record's animation setup onto Kliff, so Damian holds each target value at its real slot. Verified against the live 1.13/1.14 `characterinfo.pabgb`:

| field | block offset |
|---|---|
| `appearance_name` | +0 |
| `character_prefab_path` | +4 |
| `skeleton_name` | +8 |
| `lookup_24` | +20 |
| `lookup_25` | +24 |

`block+16` is a table-wide constant (`3938836851` across all 7,105 records) — a type tag — which is exactly where the legacy `lookup_24`→+16 mapping pointed, so that mapping wrote to a constant. The new schema routes `lookup_24` to its real slot (+20).

### Schema discriminator (keeps both worlds working)
A mod that uses the new semantic names (`appearance_name` / `character_prefab_path`) is interpreted under the new block layout; **legacy mods keep the old offsets byte-for-byte**, so nothing already in the wild changes behaviour.

### Honestly scoped
The three post-block fields the 7.6 mod also sets (`default_action_action_index`, `character_weight`, `f36`) sit in the stretch Pearl Abyss made variable-length in 1.13; their offset drifts per record and is deliberately **not** mapped — they report "could not locate" rather than a guess. So **17 of the mod's 25 intents apply** (every mesh / appearance / skeleton / animation field — the ones that make the swap work); the 8 secondary ones are skipped pending parser work on that region.

### Verification
- Byte-exact on the live table **and** a committed 6-record fixture (the five mod targets + Damian source): 17 changes, Kliff's block ends byte-identical to Damian's, the +16 type tag and every non-target record untouched.
- 9 new tests + the existing 6 characterinfo tests green; ruff clean.

Refs #302. Does not touch the separate race/gender folder-picker sub-issue in that thread, or the 8 deferred post-block fields.


AgentKush:
**Merge-order note.** Independent of #303. Overlaps #304 on one file only — `src/cdumm/engine/characterinfo_writer.py` (`_FIELD_MAP`) — so whichever of #304/#305 merges *second* conflicts there.

Trivial resolution: keep this PR's refactored 4-tuple map + `_NEW_SCHEMA_MAP`/discriminator, and re-add #304's `call_mercenary_cool_time` entry in the 4-tuple form:

```python
"call_mercenary_cool_time": ("_callMercenaryCoolTime_offset", 0, "<Q", 8),
```

AgentKush:
**Extended: the post-bool-block region is now walked, taking this mod from 17/25 to 23/25.**

The three fields this PR originally left unmapped (`default_action_action_index`, `character_weight`, `f36`) sit past the bool block in the stretch 1.13 made variable-length. Three shortcuts were tried and **disproved against the live table** before writing any code:

- a `field_schema` **tid** marker — no 4-byte candidate is unique across the target records (`0x00000F43` appears twice in Kliff and PlayerAll; every longer candidate is absent from Kliff_Clone/Kliff_AI);
- a **fixed delta** from `_isValid_offset` — correct for Damian only; Kliff reads 44288 there;
- an **8-byte run anchor** — absent from Kliff_Clone and Kliff_AI.

The region *is* walkable. The bool run is **variable length** — which is exactly why every fixed offset failed — and it's terminated by the u32 constant **900000**, present in all 7105 live records. From that anchor the layout is fixed, so the position is **computed**:

```
target = anchor + 44 + count * 4        (count read at anchor + 36)
```

The count term is what makes Kliff_AI (count=13) land correctly where a fixed delta read 4 bytes off.

**Correctness gate:** the u32 immediately before the target must be `1073741824` (f32 2.0). Every record that passes it decodes `_f36` as a clean 0..3 enum — **142/142** on live 1.15 — so a pass is treated as proof of position. A failure publishes **no offsets at all**, and the writer keeps refusing the field by name. Nothing is ever written to a guessed offset.

**Proof the position is right rather than merely plausible:** Damian is the record the mod copies, and the walker reads back exactly the two values the mod writes elsewhere — `1287066785` and `2`.

**Ceiling is verified, not assumed.** I widened the search to try *every* anchor in each record: 142 records with exactly one gate pass, 6963 with none, and **never two or more**. So there's no ambiguity to resolve wrongly, and scanning from the first anchor is already optimal. Kliff_Clone has **zero** gate-passing anchors, so its two intents genuinely can't be located, and `character_weight` stays unmapped (its slot can't be distinguished from the gate field). That's why this is **23/25 and not 25/25**.

Verified: fixture applies byte-exact with the table size unchanged; **full fast suite 2446 passed, 46 skipped, 0 failed**; ruff clean (the 3 E702 in the parser are pre-existing, confirmed by stashing).

AgentKush:
**Follow-up: `character_weight` is now mapped, and a gate I'd widened turned out to be wrong — reverted.**

Two commits on top of the walk, both driven by one new piece of evidence.

### The evidence: Character Creator **7.5** is a labelled oracle

7.6 ships as Format 3 field intents. **7.5 shipped the same edits as raw v2 offset patches — with the author's own labels attached.** That makes it ground truth for where these fields actually live:

```
offset=464     "Kliff _defaultActionActionIndex -> 2955633246"        00000000 -> A114B74C
offset=4503    "Kliff_Clone _defaultActionActionIndex -> ..."         00000000 -> A114B74C
offset=8414    "Kliff_AI _defaultActionActionIndex -> ..."            00000000 -> A114B74C
offset=43321   "PlayerAll _defaultActionActionIndex -> ..."           00000000 -> A114B74C
```

### 1. `character_weight` → mapped (documented, not inferred)

For the one record 7.6 calls `character_weight` (Kliff_Clone, value `1287066785`), 7.5's label for that same edit reads `_defaultActionActionIndex`. The other three records carry the identical value under 7.6's `default_action_action_index` **and** 7.5's identical `_defaultActionActionIndex` label. So the two names are one field renamed between DMM versions — same slot, and now mapped as such rather than dropped as an unsupported field name. It still only writes where the parser's gate has confirmed the position.

### 2. The widened gate was accepting a wrong position — **reverted**

I had replaced the "u32 before target == f32 2.0" gate with `f36 <= 3`, on the reasoning that a wrong position hitting 0..3 by chance is ~1-in-a-billion and that it's a strict superset of the old gate. Checking the walker against 7.5's labels killed that:

| record | walker | author (7.5) | |
|---|---|---|---|
| Kliff | 464 | 464 | ✅ |
| Kliff_AI | 8414 | 8414 | ✅ |
| PlayerAll | 43321 | 43321 | ✅ |
| **Kliff_Clone** | **4499** | **4503** | ❌ 4 bytes short |

The original 2.0f gate had correctly **refused** Kliff_Clone. The widened one accepted it and would have written 4 bytes into a neighbouring field. Reverted, and locked in with `test_kliff_clone_is_refused_because_the_walk_lands_4_bytes_short`.

Post-revert the walker is at **100% precision: 3 correct, 0 wrong, 1 correctly refused.**

### Verified count — measured on the live table, not the fixture

Pulled `characterinfo.pabgb`/`.pabgh` through the apply path (`_find_pamt_entry` → `_extract_from_paz`) on a live 1.15.00 install and ran the real 7.6 mod JSON through `build_characterinfo_changes`:

```
live table: body=26,431,464B  header=56,842B
index keys: 7105
records parsed: 7105 / 7105   named: 7105
mod intents: 25
RESOLVED: 23 / 25
refused: Kliff_Clone.character_weight, Kliff_Clone.f36
value mismatches: 0
```

**23/25 is the honest number** — same as previously reported here, and every one of the 23 writes carries the exact value the mod asked for. The two refusals are both the one record whose position can't be confirmed; refusing is the correct behaviour there.

CI on this branch is fully green: lint ✅ (13s), fast tests ✅ (5m32s), slow integration ✅ (10m3s). 19 characterinfo tests pass locally.

Issue #26, opened 2026-04-21
Mod Settings Menu Width


vanishdark:
Some mods with configuration do not display the checkbox due the size of the text. A resize feature for that sidebar or adjustment on the width will be nice

<img width="177" height="558" alt="Image" src="https://github.com/user-attachments/assets/bcc5d376-f006-4476-8b03-22d909a758a9" />

faisalkindi:
Thanks for the screenshot — that pinned down the exact bug.

Fixed in commit `855301e`. The checkbox indicator now sits on the **left** of the label rather than the right, so it stays visible even when the config panel slides over a narrow main window. Long labels still wrap to fit the available width, but you'll always see the toggle now and you can click it. Ships in v3.1.4.

faisalkindi:
v3.1.4 is out with the checkbox-always-visible fix. Closing this. Reopen if long labels still hide the toggle on your setup.

PR #317, opened 2026-07-27
interactioninfo: write the pivot range - Fast Pickup applied 0 of 10 intents

AgentKush:
## The bug

[Fast Pickup — Increase Range](https://www.nexusmods.com/crimsondesert/mods/2882) sets `interaction_pivot_list[0].raw_a` and `.raw_b` on five interaction records so gathering, skinning and insect catching reach further. **All ten intents applied zero bytes.**

Not a version mismatch — all ten keys are present in the live table and every `entry` name matches. `_interactionPivotList` is field #26 of `InteractionInfo` with type `None` in the PABGB schema: no descriptor. The generic walker can't reach it, and `parse_records` returns **0 records** for this table entirely.

## Identifying the field

The values decode it. The mod writes `1084227584` and `1077936128` — `0x40A00000` and `0x40400000`, the raw bit patterns of f32 **5.0** and **3.0**. DMM calls them "raw" because it carries the float's bits as an integer.

The schema names them. `InteractionPivotOverrideData` ends:

```
_interactionUpperHeight   f32    4
_targetGotoOffset         12B    (a 3-float vector)
_interactionDistance      f32    4     <- raw_a
_interactionLowerHeight   f32    4     <- raw_b
```

So when upper height is `0.0` and the goto offset is `(0,0,0)` — the common case — the pair is preceded by exactly **16 zero bytes**. That's a typed prediction from the schema, not a pattern spotted in a hex dump, and it holds: on the live 1.15 table all five targeted records resolve to exactly one such position, and it's the right one.

```
Gimmick_PickUp    2.5 / 2.5        SmallAnimal_Skin  1.5 / 2.0
Animal_Skin       1.7 / 2.0        Gimmick_Collect   2.5 / 3.0
Insect_Catch      2.5 / 2.6
```

295 of 393 records resolve uniquely. The other 98 are **refused, not guessed at** — the frame only holds where upper height and goto offset are zero, and on a record where they aren't, the scan would find some other pair of floats. Ambiguity is a refusal.

## The alignment rule, and why it isn't decoration

The first version scanned every byte for the frame. It was unique on vanilla, hit all five targets, and passed every check I had. **It was also fragile, and I only found out by round-tripping the write.**

Writing `3.0` (`00 00 40 40`) beside `3.0` produces `...00|00 00 40 40|00 00 40 40`, and a read one byte early yields `00 00 00 40` = **2.0** — inside the plausible range. So the record goes ambiguous the moment the mod is applied: it would work once and be refused on every re-apply.

On vanilla that shifted read is a denormal and gets filtered, which is exactly why the flaw stayed hidden until the patched table was fed back in. **Uniqueness that depends on the stored values is not uniqueness.**

Fields are 4-byte aligned and the frame is four u32s, so candidates sit on the run's own 4-byte grid — `run_start + 16, +20, +24 …` — and the off-by-one window is excluded by construction. `test_locator_survives_its_own_write` round-trips the write and re-locates the pair to keep it that way.

Three earlier locators were tried and thrown out, recorded in the module so they aren't retried: a fixed offset from the record's inner id-string (73 on four records, **89** on `Gimmick_Collect`); a fixed offset from the name; and the maximal zero-run end, contaminated because `2.5` is `00 00 20 40` and the run swallows the value's own leading zero bytes.

## Verified

Through the real apply path (`expand_format3_into_aggregated`) against the live 1.15 table:

```
10 intents -> 9 changes, 0 warnings
   2.5 -> 5.0   (Gimmick_PickUp, both fields: double pickup range)
   2.5 -> 3.0   1.5 -> 3.0   1.7 -> 3.0   2.0 -> 3.0   2.6 -> 3.0
```

The tenth is a correct no-op — `Gimmick_Collect`'s `raw_b` is already 3.0. Table size unchanged, 15 bytes differ, every value reads back exactly.

## Tests

`tests/test_interactioninfo_pivot_range.py` — 14 tests, re-running the derivation in CI against the committed fixture rather than trusting it: the values are float bit patterns; the locator finds the pair on all five records; the 16-zero frame is there; the locator survives its own write; the writer applies the real mod values and no-ops when already correct; six refusal paths each asserting the reason text; 295/393 pinned; every located pair is a plausible range.

Fixture added: `tests/fixtures/vanilla115/interactioninfo.pabg{b,h}.zlib` (26 KB compressed). Loaded directly with `zlib` so this PR stays independent of other in-flight branches.

Full fast suite: **2521 passed, 42 skipped**. `ruff check` (0.16.0, matching CI) clean on both added files.

Found by a full-corpus sweep — 301 Format 3 mod-files run through the real apply path against live tables.


faisalkindi:
Reviewed independently. **The pivot locator is well evidenced and I'd take it — but I'm holding on a regression: this PR silently drops three interactioninfo fields that work today.**

## The regression, reproduced end to end

Adding interactioninfo.pabgb to _WHOLE_TABLE_TARGETS at format3_apply.py:681 routes EVERY supported interactioninfo intent to the new writer, which only handles interaction_pivot_list[0].raw_a/.raw_b. Other fields reach the generic walker today and produce real bytes.

Same script, same fixture, both sides, through the real expand_format3_into_aggregated:

    field                        master   this PR
    is_blocked                   1 change  0, refused
    interaction_type             1 change  0, refused
    auto_moving_stop_distance    1 change  0, refused

On master those emit changes like 00 to 01 and 34 to 01. Here each comes back as a skip: field 'is_blocked' is not a supported interactioninfo field.

It's a loud refusal rather than corruption, so nothing gets damaged — but a mod that edits is_blocked applies today and would stop applying after this lands. That is the kind of thing that surfaces as a regression report weeks later.

The fix is the shape already used elsewhere in this dispatch: route only the pivot-list fields to the writer and let everything else fall through, rather than claiming the whole table. storeinfo and equipslotinfo do exactly that, and the comment above them spells out the reasoning — only the writer-supported list fields divert, every other intent falls through to the standard path.

Secondary drift, same area: _INTERACTION_PIVOT_RE accepts [\d+] and LIST_WRITERS registers the [] wildcard, but the writer's _PAIR_FIELDS only holds [0], so interaction_pivot_list[1].raw_a is accepted then refused. Safe direction, but the handler comment saying the writer bounds the index oversells it.

## The locator itself is good

Fixture is genuine — interactioninfo.pabgb and .pabgh extracted from the live install are byte-identical to the committed files (170,349 and 3,146 bytes).

It isn't a fixed offset but a locator that scans for a 16-zero frame followed by two f32 in [0.01, 100] on the zero-run's 4-byte grid, and requires exactly one match or refuses. Decoded values across all 295 uniquely-located records look like real data: raw_a is 1.5 on 120 records, 5.0 on 44, 2.0 on 36; named samples read sensibly (Npc_Shop_120 4.1/2.0, Dead_Loot 3.5/2.0, Gimmick_PickUp 2.5/2.5).

The strongest evidence isn't in the PR and had to be computed: every one of the 295 located pairs, all 590 floats, is an exact multiple of 0.05. A misaligned read essentially never produces that. All 295 pre-pair zero runs also have length divisible by 4, so the scan grid coincides with the real field grid on this table.

Refusal behaviour is properly conservative — candidate histogram over all 393 records is 295 unique, 8 with none, 60 with exactly two, 30 with three or more, so 98 records refuse rather than guess.

## Two smaller things

test_every_located_pair_is_a_plausible_range at tests/test_interactioninfo_pivot_range.py:192 cannot fail: it asserts every located value lies within [_MIN_RANGE, _MAX_RANGE], but locate_pivot_pair only returns positions that already passed that identical check. Harmless, but it isn't evidence — the multiple-of-0.05 property is what it should be asserting.

The docstring at interactioninfo_writer.py:40 says 291 resolve uniquely with 94 ambiguous; actual is 295 and 90, which the test's own assertion agrees with.

Also unproven, though it only affects naming rather than which bytes get written: raw_a mapping to _interactionDistance. If disk order pairwise-swaps the way BuffInfo's does, raw_a is lower-height instead.

Narrow the routing and I'll merge this — the locator work is solid and Fast Pickup applying 0 of 10 intents is worth fixing.

For transparency on my own testing: my synthetic pivot intent used an arbitrary key that isn't a locatable record, so I did not personally demonstrate a successful pivot write end to end. The writer's own 14 tests cover that and pass.

AgentKush:
You were right, and the table reproduces exactly. Thank you for catching it — routing the whole target at a writer that handles two fields was the wrong shape, and I'd copied the `_WHOLE_TABLE_TARGETS` line without copying the partition that makes it safe.

## Narrowed

The batch is now split the same way storeinfo/equipslotinfo already do it. The accept-set is the writer's own `SUPPORTED_FIELDS`, exported from `interactioninfo_writer`, so the dispatch and the writer can't drift apart. Everything else falls through to `_intents_to_v2_changes`.

Re-ran your table through `expand_format3_into_aggregated` against the committed vanilla 1.15 fixture, one intent per run, `git stash` between the two columns so it's the same script both times:

| field | PR HEAD (as you reviewed it) | narrowed |
|---|---|---|
| `is_blocked` | 0, refused | **1** |
| `interaction_type` | 0, refused | **1** |
| `auto_moving_stop_distance` | 0, refused | **1** |
| `interaction_pivot_list[0].raw_a` | 1 | 1 |
| `interaction_pivot_list[0].raw_b` | 1 | 1 |
| all five in **one** mod | 2 | **5** |

Pinned in `test_format3_whole_table_partition.py` — same file that caught this on storeinfo, since it's the same bug twice. The new cases run off the committed fixture rather than `issue_repro/`, so unlike the storeinfo one they aren't skippable.

One incidental finding while picking test values, **not** touched in this PR: `auto_moving_stop_distance` is narrow, and an out-of-range value applies zero bytes with no warning. `new=7` → 1 change, `new=12345` → 0 changes, `warnings_out` empty. That's the generic path, pre-existing and unrelated to this change, so I've left it alone rather than widen the diff. Happy to file it.

## The other four

**Regex.** Tightened to `[0]`. The locator frames one pair per record, so `[1]` had nothing to resolve to — and post-partition it would have been worse than before, because a field outside `SUPPORTED_FIELDS` now falls through to a walker that has no descriptor for `_interactionPivotList` and would apply zero bytes silently. Rejecting at validation is the loudest of the three options. Comment says that now instead of claiming the writer bounds it.

**The tautological test.** You were right that it asserted what the locator already filters on. Replaced with a property the rule says nothing about — quantisation — plus a control arm, because "100% of located values are multiples of 0.05" only means something if the surrounding bytes aren't:

```
located pairs                                        295 / 295   (100%)
every OTHER aligned float pair in the same records
that passes the same [0.01, 100.0] filter             54 /  89   (60.7%)
```

A locator landing on arbitrary bytes would look like the control group. Both numbers are asserted, so if the rule changes, both move.

**291/94.** Re-measured: **295 unique, 8 no match, 90 ambiguous** of 393. The old numbers predated the 4-byte-aligned candidate grid and were never re-measured after it. Docstring corrected, and it now says which test pins them.

**`raw_a` → `_interactionDistance`.** Softened. The 16-zero frame and the two f32 slots after it are measured; which of the two slots carries which name is inferred from declaration order, and declaration order isn't wire order in this format — BuffInfo pairwise-swaps. Nothing depends on getting it right (`raw_a` writes +0, `raw_b` +4 regardless, which is DMM's mapping, so a mod round-trips either way), but it was stated harder than the evidence. The docstring now says which part is measured and which is inferred.

Fast suite green, ruff clean on the files this PR adds, and the two modified engine files carry the same 40 findings as before the change — no new ones.

AgentKush:
Follow-up. Your review found a test asserting what the rule already guarantees, so I went looking for the same class of problem elsewhere and measured coverage **of the diff** rather than trusting "suite green".

```
                        added executable lines   never executed
before                          154                   20   (87.0%)
after                           154                    0  (100.0%)
```

All 20 were error and refusal paths — the parts that only run when something has already gone wrong, which is exactly where an untested branch does damage:

- `locate_pivot_pair`'s three early-outs (record shorter than the envelope, `name_len` nonsense, name running past the record end). A truncated record has to refuse, not index into whatever follows it.
- the **builder** acting on a locator refusal. The locator's refusal was tested in isolation; the builder turning that into a dropped intent with a reason was not.
- that refusal reaching `warnings_out` rather than only the log. Refusing quietly is the failure mode this writer exists to prevent, and nothing asserted it didn't.
- the `except Exception` guard — a bug in one table's writer must not abort the other mods in the same Apply. Driven now with a monkeypatched writer that raises, asserting the standard-path intent still lands.
- `participating_mod_ids` attribution on **both** sides of the partition. Missing it on either side means a mod isn't credited for bytes it actually applied.

Two commits earlier in this same push also covered the two behaviour changes from the last round that I'd shipped without tests — the `[0]`-only regex and the suppressed "0 byte changes" warning. Both measured first: `[1]` and `[7]` are rejected at validation, and the warning is silent when the walker contributed while still firing when nothing applied at all. The regex test also asserts the apply dispatch's accept-set agrees with validation, so they can't drift into "validated but never routed".

Mutation-checked rather than assumed:

```
revert the regex to \d+              -> 2 tests fail
drop the pt_changes guard            -> 1 test fails
revert the partition entirely        -> 5 tests fail
narrow the crash guard to ValueError -> 2 tests fail
```

No behaviour changed in this pass — it only means the claims I was making are now claims a test defends.

AgentKush:
Heads-up: this was `CONFLICTING` against master and I hadn't noticed — #316 landed on the same lines (`_WHOLE_TABLE_TARGETS` and the adjacent whole-table dispatch blocks). Merged master in and it's `MERGEABLE` again.

Resolved by taking master's `format3_apply.py` / `format3_handler.py` wholesale and re-applying this branch's three hooks on top — the target-set entry, the dispatch block, and the regex with its `_classify_intent` early-accepts — rather than hand-untangling interleaved dispatch blocks. Positional conflict only, no semantic change on either side.

Re-ran the tests and re-measured diff coverage **on the merged base** rather than carrying the earlier numbers forward, since a green result on a stale base doesn't transfer: still 0 uncovered added lines.

Worth knowing if you merge both: #317 and #320 now conflict with **each other**, for the same reason — they insert their dispatch block at the same anchor. Whichever goes second needs the same trivial resolution (keep both blocks, union the target set). Happy to rebase the second one onto the first once you've picked an order.

PR #319, opened 2026-07-27
feat(statusinfo): apply stat_level_data mods at the measured offset

AgentKush:
Brings the DIRECT SPEED stat writer back, on an offset and encoding an oracle settled rather than a third guess at a constant.

Refs #304. Refs #191 (the DIRECT SPEED stat-mod support promised in that thread lives here now, not in #304).

## Where this left off

You pulled this out of #304 because `_SLD_TAIL_OFFSET = 80` ran the 16-element array off the end of the tail into an 8-byte value at +204 that is byte-identical on all four rate records and never zero. You were right, and the tests could not have caught it — their containment bound came from the same `+80`.

You proposed 76 and left the encoding open, because vanilla read as 32.32 gives `MoveSpeedRate[1] = 78125.0` against a mod writing `40000000` — 512× as a value, 1/8400 as a raw u64, and "2x move speed" either way it's read.

**76 is not the offset either**, and nothing in the table separates it from 68: both partition the 212 bytes exactly, and *any* consistent scale produces a tidy ramp, because 2²⁴ and 2³² differ by a constant factor. The question isn't answerable from the table — which is why two offsets have now shipped wrong.

## The oracle

Mod 2511 publishes, for the **same game version (1.13.0)**, both the JSON V3 field mod (file 12184) *and* a PAZ overlay carrying the already-modded `statusinfo.pabgb` (file 12185). The request and its result ship side by side, so the diff answers offset and encoding outright.

Searching every offset that partitions the tail into whole u64s, then keeping only those that contain every byte the overlay actually changed, leaves exactly two candidates — 68 and 76:

| offset | `raw == asked << 24` |
|---|---|
| **68** | **144 / 144** |
| 76 | 0 / 144 |

```
_SLD_TAIL_OFFSET = 68
16 elements of uint64
raw = value << 24          (24-bit fixed point)
```

That accounts for the whole 212-byte tail — a 68-byte header, the array at 68..195, a zero terminator slot at 196, then the constant at 204. Which is precisely what 80 was overrunning into.

## Verification

The mod's own intents, run through the **full pipeline** (DB row → `expand_format3_into_aggregated` → whole-table dispatch → writer → applied), reproduce the author's shipped table **byte for byte** on all six presets:

```
2xAtkSpd                 1 change   BYTE-EXACT
2xAtkSpd_2xMovSpd        2 changes  BYTE-EXACT
3xAtkSpd                 1 change   BYTE-EXACT
3xAtkSpd_2xMovSpd        2 changes  BYTE-EXACT
4xAtkSpd                 1 change   BYTE-EXACT
4xAtkSpd_2xMovSpd        2 changes  BYTE-EXACT
```

Zero residual differences, zero warnings, zero refusals.

The mod files can't be committed, so the tests carry corroboration that lives inside the fixture instead:

- All **64** vanilla elements decode to exact whole numbers on monotonic ramps — DHIT `0, 2000, 4000 … 25000`; AttackSpeedRate `0, 29000000, 55000000 … 250000000`.
- The presets are what their names say: **2xAtkSpd sets level 1 to 58,000,000 against a vanilla 29,000,000 — exactly double.**

Neither survives shifting the offset by one slot in either direction, so `test_vanilla_decodes_to_exact_whole_ramps` is a real pin rather than a restatement of the constant. `test_trailer_and_terminator_are_never_touched` pins the specific defect 80 had.

## Refusals

The guardrails are what stop a wrong offset being silent next time:

- a record whose tail isn't 212 bytes (the 71 regular stats have no `stat_level_data` at all)
- an index outside 0..15, which would reach the terminator or the constant
- a value that doesn't fit once shifted, or isn't a whole number — including `bool`, which is an `int` subclass and would otherwise be written as `1`

Each is returned to the caller and surfaced to the user; nothing is dropped in silence.

This also closes the `_pack_i64` defect you found: `stat_level_data[0] = 10000000000` used to store raw and read back `1410065408` with `dropped == []`. It now round-trips whole, and `test_largest_representable_value_is_accepted` pins the refusal to the exact field boundary rather than somewhere below it.

Writes are length-preserving, so there's no companion `.pabgh` rebuild.

## Checks

- Fast suite: **2475 passed, 42 skipped** (2452 on `master` + the 23 new)
- `ruff` clean on the added files; **zero new findings** on the two modified ones

## One dependency worth flagging

Reading that overlay at all needed a separate extraction fix — `decompress_entry`'s stored-uncompressed path returned ciphertext untouched, so the table came out as 8185 of 8216 bytes of noise. That's #318, independent of this and not required to merge it.

Thanks for holding the merge and for not patching the constant yourself. Had you done it, we'd have shipped 76 and it would have been wrong in a way the tests couldn't see — the same shape as the 80 defect, one step over.


AgentKush:
Independent corroboration turned up while working a different table, and it's worth adding because it's the first evidence here that isn't a byte pattern.

Nexus mod **2634 "Critical Rate Enhancement"** by *norva2* — different author, different mod, game version **1.08** — ships a statusinfo field mod whose intents carry human-readable notes stating the vanilla values:

```json
{ "field": "stat_level_data[0]",  "new": 0,
  "note": "Critical Rate Level 0 - default : 0" }

{ "field": "stat_level_data[15]", "new": 500000000,
  "note": "Critical Rate Level 15 - default : 250000000" }
```

Decoding the committed 1.13 fixture at offset 68 with a 24-bit shift gives CriticalRate as:

```
0, 29000000, 55000000, ... 240000000, 250000000
^                                     ^
level 0 = 0                           level 15 = 250000000
```

Both match exactly. A third party, on a different game version, wrote down in plain text what the field holds — and it agrees with the decode to the digit. At 76 or 80 it would not.

The same mod is also a second, independent instance of the doubling semantics: level 1 goes `29000000` → `58000000`, same as 2xAtkSpd, from a different author who called it "Double Critical Rate".

No code change — just closing the loop on the encoding question.

PR #320, opened 2026-07-27
feat(statusgroupinfo): apply status_info_list writes

AgentKush:
Nexus mod 2634 "Critical Rate Enhancement" (norva2) ships a Global Critical Rate patch that retargets one slot of the item-activation stat groups:

```json
{ "key": 1000006, "field": "status_info_list[3]", "op": "set", "new": 1000006 }
```

`statusgroupinfo` has no CDUMM schema, so those intents were refused with a generic "unsupported field" and the mod did nothing.

## The record grammar

```
<u32 key><u32 name_len><name><u8 is_blocked>
<u32 c><c * u32>    x3     three lists of statusinfo record keys
<u32 75><75 * u32>  x2     two reverse index tables
```

Established by **exact tiling** — the walk consumes all eight records to the byte, nothing short and nothing left over, which a wrong field order or element width cannot do.

75 is exactly the number of `statusinfo` records, and the two tables are reverse indexes. Each carries one set entry (the rest are `0xFFFFFFFF`) per element of the list it serves, and the entry value is that element's position in the list — a permutation, and checked as one. That correspondence holds on all eight records for both tables, which is what pairs table 0 with list 1 and table 1 with list 0. The slot ids are stable across records: slot 9 resolves to CriticalRate wherever it's set, slot 10 to DHIT, slot 4 to DDD.

## Which list is `status_info_list` — and why the writer doesn't just trust the answer

The shipped schema (`schemas/pabgb_complete_schema.json`, `StatusGroupInfo`) names three list-typed fields — `_statusInfoList`, `_regenerateStatusInfoList`, `_elementalStatusInfoList` — and the wire has exactly three lists. Its **declaration order does not settle the mapping**: that order is demonstrably not wire order for this table, since it puts `_regenStatusIndexList` before `_key` and swaps `_isBlocked` with `_stringKey`.

The contents settle it, the same way on both record families:

- **List 2** is `[Temperature, Confusion, Electricity]` on every record that has one — the three elemental statuses.
- **List 0** holds Hp, Fatal, Hunger, Stamina, Mp, Morale, Fury. **List 1** holds DDD, DPV, DHIT, CriticalRate, AttackSpeedRate, the resistances.
- On `StatOnActivateByItem`, whose whole purpose is the stats an item grants, **lists 0 and 2 are empty** and list 1 carries the stats. A group named for granting stats cannot have an empty `_statusInfoList`.

That's an inference from content, not a byte-level proof — so the writer doesn't lean on it alone. **An intent is refused whenever another list in the same record is also long enough to hold the index**, which is precisely the case where a mis-naming could do damage. On the two records this mod targets the other lists are empty, so the answer is forced whatever the naming is; on the six numeric groups every list is populated and the writer refuses rather than pick.

If you can confirm the naming against the upstream tool's schema, the fix is one constant (`_STATUS_INFO_LIST`).

## Refusals

- a record that does not tile under the grammar — a table whose layout has drifted is left alone, not written on stale offsets
- an index no list can hold
- an index more than one list could hold
- a value that isn't a 32-bit record key, including `bool`

## Verification

The mod's own file through the full pipeline (DB row → `expand_format3_into_aggregated` → whole-table dispatch → writer → applied):

```
changes: 2   warnings: []   length preserved: True   bytes changed: 2

key 1000006  slot 3: 1000007(CriticalRate) -> 1000006(CriticalDamage)
key 1000007  slot 3: 1000007(CriticalRate) -> 1000006(CriticalDamage)
```

- Fast suite: **2471 passed, 42 skipped** (2452 on `master` + the 19 new)
- `ruff` clean on the added files; **zero new findings** on the two modified ones

## One thing worth flagging

The mod's own effect is a little odd — it replaces CriticalRate with CriticalDamage in a group that already contains CriticalDamage, which isn't obviously "global critical rate". That's the author's call, not ours; CDUMM's job is to apply what the file says. Noting it so it isn't mistaken for a bug in the writer if someone tests the mod in-game.


faisalkindi:
Reviewed independently in an isolated worktree. **The layout is proven — more strongly than the PR claims — but I'm holding this one on a different question: the write leaves the reverse index stale.**

## The layout is right, and the evidence is better than what you cited

Decoded the committed 1.15 fixture with an independent walker rather than the PR's. Four confirmations:

- All 8 records tile exactly under the grammar at statusgroupinfo_writer.py:88-124.
- Every decoded value resolves to a real stat name via the repo's separate stat_names.py snapshot. Key 1000006 StatOnActivateByItem list1 = DDD, DHIT, CriticalDamage, CriticalRate, AttackSpeedRate, GuardPVRate. No garbage anywhere.
- The reverse-index tables form a global bijection across the whole table: every occupied slot maps to exactly one statusinfo key, zero conflicts. A wrong field order or element width could not produce that.
- A naming oracle the PR doesn't mention: key 1000007 StatOnActivateByItemWithoutAttackSpeedRate has list1 = 1000006's list1 minus AttackSpeedRate. The record's own name describes list1's contents, which independently confirms _STATUS_INFO_LIST = 1.

Refusal behaviour is right too: parse_record returns None unless the walk lands exactly on the record end, and refusals reach the user through warnings_out rather than only the log. Length-preserving, 6230 to 6230 bytes, exactly 2 bytes changed. Reachability traced through apply_engine.py:798 to the dispatch at format3_apply.py:1132 by driving the real pipeline, not by reading.

## The problem

table0 is a reverse index over list1 — the PR establishes this itself and asserts it in test_the_two_index_tables_pair_with_the_lists. The write then changes list1 without touching table0.

I checked the invariant before and after the mod's write:

    VANILLA  occupied slots, 0 conflicts
    PATCHED  same slots, conflicts introduced

A slot still claims CriticalRate (1000007) while the list position it points at now holds CriticalDamage (1000006).

The PR's own test can't catch this, because it asserts only the weak form: entry count equals list length, and values are a permutation of range(len). Both survive the write untouched. The strong form, slot id to record key, is what breaks, and nothing asserts it post-write.

And the harder part: the mod's edit isn't cleanly representable here. It puts CriticalDamage at both positions 2 and 3 and removes CriticalRate entirely. A reverse index maps one slot to one position, so there is no correct table0 for that list. Setting the slot to 0xFFFFFFFF would then break the permutation property. This isn't the PR forgetting to update the index — the requested edit produces a state vanilla never exhibits.

## What I can't settle, and why I'm not merging on a guess

Whether the game reads the list, reads the reverse table, or rebuilds the index at load. That determines whether the stale slot is harmless, silently wrong, or a crash. The PR's verification is pipeline-level only, and I have no in-game evidence either way. Writing a state the game never ships, without knowing which structure it trusts, is the one thing this repo consistently refuses to do.

Worth checking whether DMM updates table0 when it applies this mod. If it doesn't and the mod demonstrably works in game, that settles it and I'll merge immediately.

## Two cheap things regardless

new is never checked to be a real statusinfo key — statusgroupinfo_writer.py:158 only checks it's a 32-bit int, so status_info_list[3] = 99999 is accepted and written. Valid keys are 1000000 to 1000074, and the reverse-index tables already hand you a 75-slot key space to range-check against. A dangling record reference in a list the game dereferences is a plausible crash vector.

The ambiguity guard also covers less than the commit implies: it only fires when more than one list can hold the index, so above len(list0) only list1 fits and the naming assumption is load-bearing and unguarded across roughly a hundred index positions. My evidence says the naming is right, so that's residual risk rather than a demonstrated bug, but the claim that the answer is forced whatever the naming is holds only for the two target records.

Tests: test_statusgroupinfo_writer.py 19 passed; format3/statusgroup sweep 397 passed, 2 skipped. Ruff clean on the new files; the 39 findings in the modified ones are identical on HEAD~1.

faisalkindi:
Follow-up on my hold. I went looking for the same kind of oracle that settled the statusinfo offset in #319, since if a statusgroupinfo-touching mod shipped a Mod Folder Version overlay we could just read whether table0 gets updated. **There isn't one reachable here.**

Searched the fixtures, issue_repro, every local .pamt and .paz overlay, the CDUMM mod databases (both empty), and the local mod-pack corpus. Zero already-modded statusgroupinfo anywhere. Mod 2634 isn't present locally in any form.

One positive result worth recording, because it closes the last doubt about the layout proof: I extracted gamedata/statusgroupinfo.pabgb and .pabgh from the live Steam install and diffed them against the fixture this PR commits.

    statusgroupinfo.pabgb: live 6230B, fixture 6230B, identical
    statusgroupinfo.pabgh: live 66B,   fixture 66B,   identical

So the fixture is genuine game data, and the layout evidence in my earlier comment rests on the real table. The layout is not in question. It also confirms the install is vanilla for this table, so there's no applied overlay to read there either.

For completeness: the only file on this machine containing the string status_info_list is a DMM tool binary under issue_repro. I did not extract from it and won't. Published mod output is a fair source; pulling field layout out of the competing tool's compiled binary is exactly what its license restriction targets. Mentioning it so you know the lead was seen and deliberately left alone rather than missed.

That leaves the original question needing someone to launch the game with this patch applied. Worth adding: even a recovered overlay might not settle it, because the mod's edit puts CriticalDamage at both list positions 2 and 3 and removes CriticalRate entirely, and a one-slot-to-one-position reverse index has no correct representation for that. An overlay would show whatever the authoring tool emitted, which is evidence about that tool rather than about what the game requires.

So: the layout is proven and I'm happy with it. The hold is purely about writing an index state vanilla never ships, without knowing which structure the game trusts. If you've run this in game and it behaves, say so and I'll merge on that.

AgentKush:
Thank you for both of these — the independent decode, the naming oracle I'd missed, and the live-install diff of the fixture are all more than I'd asked for, and the hold is a real finding I should have caught.

## I reproduced your finding before responding to it

Re-derived the reverse index from the committed fixture with a walker written from the grammar rather than by importing `parse_record`, so it isn't the PR marking its own homework:

```
VANILLA   53 occupied slots, 53 distinct keys, 0 conflicts   (bijection)
PATCHED   1 conflict per edited record
          slot 9 claims 1000007 (CriticalRate) but the position
          it points at now holds 1000006 (CriticalDamage)
```

You're right, and the PR's own test can't see it — `test_the_two_index_tables_pair_with_the_lists` asserts count-equals-length and permutation-of-range, and both survive the write. Also confirmed the slot map is a real permutation, not arithmetic: 1000006 → slot 12, 1000007 → slot 9, 1000010 → slot 7. It has to be read from the table.

## One thing I can settle, which narrows your three outcomes to two

You framed it as harmless / silently wrong / crash. A crash needs a pointer that dangles or leaves its list. I measured whether the write produces one — across the whole table, both index tables, after applying both of the mod's intents:

```
every record still tiles exactly       True
index entries checked                  262
out-of-range pointers                    0
pointers at an absent element            0
length                          6230 -> 6230
bytes changed                              2
```

Nothing is removed, no list changes length, no slot is freed or added, and every occupied slot still points in range at a real element. The only thing that changes is which key the element at that position *is*:

```
StatOnActivateByItem
  before: [DDD, DHIT, CriticalDamage, CriticalRate,   AttackSpeedRate, GuardPVRate]
  after : [DDD, DHIT, CriticalDamage, CriticalDamage, AttackSpeedRate, GuardPVRate]
  slot 9 (CriticalRate) now resolves to CriticalDamage, position 3, in range
```

So it's semantic staleness, not structural corruption. Walking the three plausible read strategies:

- game reads `list1` directly → sees CriticalDamage twice, no CriticalRate. Mod works.
- game reads `table0[slot(X)] → list1[pos]` → CriticalRate resolves to CriticalDamage. Mod works, and this is exactly what it's asking for.
- game rebuilds its own map from `list1` at load → CriticalDamage appears twice; whichever wins, both entries are the same key. Consistent either way.

I can't turn that into "harmless" — that's still an argument from structure, not a launch — but I don't think crash is on the table any more, and I've pinned all of it in tests rather than leaving it in this thread:

- `test_vanilla_reverse_index_is_a_global_bijection` — 53 slots, 0 conflicts
- `test_the_mods_write_leaves_the_reverse_index_semantically_stale` — asserts your finding, `[(9, 1000007, 1000006)] * 2`
- `test_the_stale_slot_is_semantic_not_structural` — in-range on every entry, still tiles, length-preserving

I deliberately wrote the second one to assert the *stale* state rather than paper over it. If someone later makes the writer maintain the index, that test is the one that has to change, and changing it forces them to say so.

## Why I didn't just make it maintain the index

I costed it. The slot map is derivable (bijective, 53/53), so for a write `list1[i] = K` you can set `table0[slot(K)] = i` and clear `table0[slot(old)]`. That works for 4448 of the possible single-element writes across the table.

It does not work for this mod's. Both its intents put a key that's already in the list at a second position — record 1000006's list1 already contains 1000006 at position 2 — and a one-slot-to-one-position index has no representation for that, exactly as you said. The mirror trick (leave `list1` alone, point slot 9 at position 2) just breaks the bijection in the other direction, two slots on one position.

So maintaining the index would refuse this mod rather than fix it, and it'd ship ~4.4k code paths no mod currently exercises. That felt like the wrong trade to make unilaterally while your hold is open — say the word and I'll add it, but I'd rather not guess at your preference.

## The two cheap ones are done

**Key-space range check.** `new` is now bounded to the statusinfo key space, so `status_info_list[3] = 99999` is refused with a reason instead of written. Not hardcoded: it comes from the repo's own `stat_names.STAT_NAMES_CD113` snapshot, and `test_the_key_space_bound_agrees_with_the_index_table_width` pins that the snapshot has exactly `_TABLE_LEN` keys — 75 statusinfo records, 75 reverse-index slots. If those ever disagree the bound is wrong and the test says so. There's a fallback to the literals if the snapshot module isn't in a trimmed build, so it range-checks either way.

**The ambiguity-guard overclaim.** You're right that "the answer is forced regardless of the naming" only holds for the two records the mod touches, and only because their list 0 and list 2 are empty. The docstring now has a section headed *How far the ambiguity guard actually reaches* that says the guard only fires on overlapping indexes, that above `len(list0)` the naming inference is load-bearing and unchecked across roughly a hundred positions, and that the content argument plus your two oracles are what carry it there. The inline comment at the guard points at that section instead of implying more.

## Where that leaves it

The hold stands as far as I'm concerned — I have no in-game evidence, and I'm not going to claim structure-implies-safe. I don't have DMM's applier behaviour either, and I agree the binary is off limits; I'd add that even a recovered overlay would only tell us what its authoring tool emits, which is your point too.

Happy to leave this open until someone runs it. If you'd rather have the safe subset now, I'll flip it to maintain-the-index-and-refuse-duplicates and this mod becomes an explicit refusal with a reason instead of a silent no-op — your call, and it's a small change from here.

`test_statusgroupinfo_writer.py` 30 passed. Ruff clean on both files.

AgentKush:
Follow-up. After the last round I went back and measured coverage **of the diff** rather than trusting "suite green" — an added line no test executes is a line shipped on reasoning alone, which is the same failure mode as the tautological test you caught.

```
                        added executable lines   never executed
before                          155                   47   (69.7%)
after                           155                    0  (100.0%)
```

The writer was the well-covered part. The gap was the **dispatch that connects it to Apply** — 33 of the 34 added lines in `format3_apply.py` never ran under test. You traced reachability by hand in your review; there was no test doing it. Four now drive the real pipeline: the mod's actual intent end to end (2 records, 2 length-preserving changes, `_target_file` set, mod attributed), a refusal surfacing in `warnings_out` with its reason, the "0 byte changes" warning on a no-op, and the `except Exception` guard driven with a writer that raises.

Writer guards that had never executed: `parse_record` on malformed envelopes, `_read_list`'s sanity bound, the 75-entry table width check, both tiling arms, a non-`status_info_list[N]` field, and a non-integer key.

Two of those are worth calling out because they were quietly wrong rather than merely missing:

**The "did not tile exactly" arm needed the record end *extended*, not shrunk.** Shrinking trips the width check first, so the obvious test exercises a different branch and the tiling refusal looks covered when it isn't. That's the arm the whole refuse-rather-than-guess contract rests on.

**A short `.pabgh` does not raise.** `parse_pabgh_index` warns and returns no offsets, so intents fall through to "no record with key" — the `header unreadable` guard is for the raising case only. It's monkeypatched now, and both routes have a test saying which is which.

All mutation-checked rather than assumed:

```
drop the exact-tile refusal        -> 1 test fails
dispatch stops surfacing refusals  -> 1 test fails
writer stops range-checking        -> 8 tests fail
```

None of this changes the writer's behaviour, and none of it touches your hold — that still needs someone to launch the game. It just means the parts I *was* claiming are now claims a test defends.

AgentKush:
Heads-up: this was `CONFLICTING` against master and I hadn't noticed — #316 landed on the same lines (`_WHOLE_TABLE_TARGETS` and the adjacent whole-table dispatch blocks). Merged master in and it's `MERGEABLE` again.

Resolved by taking master's `format3_apply.py` / `format3_handler.py` wholesale and re-applying this branch's three hooks on top — the target-set entry, the dispatch block, and the regex with its `_classify_intent` early-accepts — rather than hand-untangling interleaved dispatch blocks. Positional conflict only, no semantic change on either side.

Re-ran the tests and re-measured diff coverage **on the merged base** rather than carrying the earlier numbers forward, since a green result on a stale base doesn't transfer: still 0 uncovered added lines.

Worth knowing if you merge both: #317 and #320 now conflict with **each other**, for the same reason — they insert their dispatch block at the same anchor. Whichever goes second needs the same trivial resolution (keep both blocks, union the target set). Happy to rebase the second one onto the first once you've picked an order.

AgentKush:
Rebased onto current master — was CONFLICTING, now clean.

Master gained the interactioninfo whole-table block (#317) at exactly the spot this branch inserts the statusgroupinfo one, and the two blocks end in identical tails. Git's textual merge interleaved them into something that still compiled but dispatched neither correctly, so I didn't hand-resolve the hunks.

Both are self-contained `if target == "...": ... continue` blocks, so the correct resolution is "keep both in sequence". I rebuilt `format3_apply.py` from master verbatim and inserted this branch's block whole, ahead of the inventory dispatch, with `_WHOLE_TABLE_TARGETS` carrying both entries. Same shape for `format3_handler.py`: master's `_INTERACTION_PIVOT_RE` kept as-is, this branch's `_STATUSGROUP_SIL_RE` and its `_classify_intent` hook re-added on top.

Verified the two coexist rather than assuming it: 414 passed, 2 skipped across the statusgroup / interaction / partition / format3 selection, and the full fast suite after that.

The in-game hold from your last round still stands — nothing here changes the writer, only where it sits relative to #317.

AgentKush:
Fixed. The index now travels with the write, and the cases where it can't are refused.

## What changed

`table 0` slot `s` holds the **position in that record's list 1** where slot `s`'s key sits, or `0xFFFFFFFF` when the record doesn't carry it. So `list1[table0[s]]` names the key that slot stands for, and reading that across every record recovers the whole correspondence. That's `learn_slot_keys`, and it returns `None` if any two records disagree rather than handing back a half-built map.

It has to be read, not computed — slot 3 is key 1000074 while slot 4 is key 1000002. Your global-bijection finding is what makes it recoverable at all: 8 records, 53 occupied slots, no slot naming two keys, no key claiming two slots.

## A representable write now keeps the invariant

Substituting a key that isn't already in the list touches three u32s — the list element, the slot the new key needs, and the slot the displaced key vacates:

```
VANILLA  184 occupied slots, 0 conflicts
PATCHED  184 occupied slots, 0 conflicts     slot->key map unchanged
         list1[3]  1000007 -> 1000011
         slot 9    (CriticalRate)  -> 0xFFFFFFFF
         slot 8    (1000011)       -> 3
         length preserved
```

## The mod is refused, and that's the answer

You were right that its edit isn't cleanly representable — CriticalDamage at 2 and 3, CriticalRate gone. A reverse index maps one key to one position, so there is no correct `table 0` for that list. So the writer refuses it:

```
writing status_info_list[3] = 1000006 on record 1000006 would list [1000006]
more than once; table 0 is a reverse index over this list and cannot point
one key at two positions, so there is no consistent record to write
```

Surfaced through `warnings_out`, verified end-to-end through the real pipeline. **Mod 2634 no longer applies** — that's a deliberate regression against the PR's original goal. I'd rather ship a correct `status_info_list` writer that declines one unrepresentable edit than one that writes a state vanilla never exhibits and hopes the game reads the list. That also means the question you couldn't settle — harmless vs silently wrong vs crash — no longer has to be answered before merging: nothing writes that state now.

Also refused: a key the index has no slot for, since the slot that must point at it is unknown. And intents are batched per record first, because two edits that each look fine can together duplicate a key.

## Your two cheap things

Both were already in from the last round: `new` is range-checked against the statusinfo key space (1000000-1000074, cross-checked against `_TABLE_LEN` at import, with a fallback so a trimmed build still checks), and the docstring now states the ambiguity guard's actual reach instead of claiming the answer is forced regardless of naming.

## Tests

The two tests that documented the bug now prove the fix — `test_the_mods_write_leaves_the_reverse_index_semantically_stale` became `test_a_representable_write_leaves_the_index_a_bijection`, asserting slot->key is unchanged after the write. The weak assertion you called out is now joined by the strong one post-write, plus the displaced-slot half, plus a test that the map is learned rather than assumed.

54 passed in the statusgroupinfo suite; 590 passed / 4 skipped across the format3, statusinfo and handler sweep. Ruff clean on both new files; `format3_apply` 25 -> 25 and `format3_handler` 17 -> 17, unchanged.

Rebased onto current master — the conflict was #316/#317's whole-table blocks landing on the same lines again, so I took master's copies of both dispatch files and re-applied this branch's three hooks rather than hand-resolving interleaved hunks.

---

_(Reposting: the previous version of this comment had its backticks mangled by a shell-quoting mistake on my side, which broke the code fences. Same content, correctly formatted.)_

PR #322, opened 2026-07-28
variant re-import: stage first, swap last (Fixes #314)

### LINKED ISSUE #314: variant re-import: a failure mid-way deletes the existing mod's files with no rollback
Found while reviewing #303. Pre-existing in `import_multi_variant`, but #303 newly routes zip / 7z / CLI imports through that function, so the exposure is wider than before.

## The problem

In `src/cdumm/engine/variant_handler.py`, a re-import destroys the old state **before** writing the new state:

- `DELETE FROM mod_deltas` (~line 544)
- `shutil.rmtree(old_dir)`
- ... then `copy_variants_to_mod_dir` (~line 576)

If anything fails between the delete and the copy — disk full, a locked file, an antivirus hold on a freshly written file, a permissions blip — the user's existing mod files are already gone and nothing puts them back.

The new caller wraps the whole thing in a broad handler (`import_handler.py`, the `except Exception` around `import_multi_variant`) and returns `None`, so the caller falls back to the generic "drop it on the main window" message. From the user's side it reads as though nothing happened, while their installed mod has actually been deleted.

## Why it matters

This is the same failure shape the transactional PAZ work was built to avoid: never destroy the old thing until the new thing is known good. Re-import is the common path for updating a mod, so this runs often.

## Suggested fix

Stage first, swap last:

1. Copy the new variants into a temp directory alongside the target.
2. Only once that succeeds, delete the old directory and rows.
3. Move the staged directory into place.
4. On any failure, remove the staging dir and leave the existing mod untouched.

Alternatively, rename the old directory aside rather than deleting it, and only remove it after the copy succeeds — cheaper, and still recoverable.

## Acceptance

- Inject a failure in `copy_variants_to_mod_dir` on a re-import; the previously installed mod must still be present and enabled afterwards, with its `mod_deltas` rows intact.
- The user-visible message on that path should say the update failed and the existing install was kept — not the generic drop-it-on-the-window text, which is misleading here.

## Related

The other #303 review finding, the variant pick being silently reset on re-import, is already fixed on master in c669132.

AgentKush:
Fixes #314. Your diagnosis and your suggested fix, implemented.

## Reproduced first

Running the new tests against master's `variant_handler`, the installed mod directory comes back **empty**:

```
E  AssertionError: installed files must survive
E  assert {} == {'merged.json': ...,
E                'variants/alpha.json': ...,
E                'variants/beta.json': ...}
E  Right contains 3 more items
```

All three files gone, and `import_multi_variant` raised into a caller that reports "couldn't import" — exactly the shape you described.

## Stage first, swap last

Everything is built in a sibling staging directory. The old content isn't touched until the replacement is complete and verified, the swap is a rename, and the previous directory is kept until the database has committed — so any failure along the way can put it back.

## The part I'd flag for review

The obvious guard — *"if we already swapped, restore"* — is wrong, and it's wrong in the same direction as the original bug.

There are **two** renames. If the second one fails, the old directory has already been moved aside while `swapped` is still `False`. Keying the restore off `swapped` sends that case down the "nothing was touched" branch, and the cleanup then deletes the backup — losing the mod, from inside the code meant to protect it. I wrote it that way first; `test_a_failure_during_the_swap_restores_the_previous_directory` is what caught it.

The restore keys off the backup existing instead, which is true in both failure shapes.

## Verification, not just staging

While in here: `copy_variants_to_mod_dir` logs and continues when a copy fails, so it returning is **not** evidence the files arrived. A half-populated staging directory swapped over an installed mod loses variants silently — same class of problem, quieter failure mode.

`verify_staged_variants` checks every expected file is present and non-empty and raises `VariantStagingError` naming what's missing, before anything is swapped. The caller already handles a raised exception, so the user gets "couldn't import" instead of a mod that looks fine and is missing half its variants.

Say the word if you'd rather that stayed out of this PR — it's a separate defect, just one that makes the staging guarantee real rather than nominal.

## Also

The database work is now rolled back on failure rather than left pending on the connection, where an unrelated later commit could have flushed a half-applied re-import.

## Tests

Seven, each failing the import at a different point and asserting the same two things: the previously installed files are still there, and the database still describes them.

| test | fails at |
|---|---|
| `a_failed_copy_leaves_the_installed_mod_untouched` | staging (disk full) — old files never touched |
| `a_silently_incomplete_copy_is_caught_before_the_swap` | staging returns with a variant missing |
| `a_failure_during_the_swap_restores_the_previous_directory` | between the two renames |
| `no_staging_or_backup_directories_are_left_behind` | success path cleanup |
| `a_successful_reimport_replaces_the_content` | stale files don't survive the swap |

Three of them fail on master; the other four pass there and are regression cover.

## Checks

- Fast suite: **2459 passed, 42 skipped** (2452 on `master` + the 7 new)
- `ruff` clean on the added test; **zero new findings** on `variant_handler`

### DIFF
--- a/src/cdumm/engine/variant_handler.py
+++ b/src/cdumm/engine/variant_handler.py
@@ -41,7 +41,10 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import shutil
+import tempfile
+from contextlib import suppress
 from pathlib import Path
 
 logger = logging.getLogger(__name__)
@@ -328,6 +331,48 @@ def build_variants_metadata(
     return variants
 
 
+class VariantStagingError(RuntimeError):
+    """A staged re-import didn't produce everything it promised.
+
+    Raised before anything the user already has is touched, so the
+    caller can abort with the installed mod still intact.
+    """
+
+
+def verify_staged_variants(
+    presets: list[tuple[Path, dict]], staged_dir: Path,
+) -> None:
+    """Check the staged copy actually contains every variant.
+
+    ``copy_variants_to_mod_dir`` logs and continues when a copy fails,
+    so "it returned" is not evidence that the files are there. A
+    re-import that swaps a half-populated directory over the user's
+    installed mod would lose variants silently, which is the same
+    class of problem as GitHub #314 with a quieter failure mode.
+
+    Raises :class:`VariantStagingError` naming what's missing.
+    """
+    expected = _disambiguate_basenames(presets)
+    vdir = staged_dir / "variants"
+    missing = [
+        expected.get(src, src.name) for src, _data in presets
+        if not (vdir / expected.get(src, src.name)).is_file()
+    ]
+    if missing:
+        raise VariantStagingError(
+            f"{len(missing)} of {len(presets)} variant file(s) did not "
+            f"copy: {', '.join(sorted(missing)[:5])}"
+            + (" ..." if len(missing) > 5 else ""))
+    empty = [
+        expected.get(src, src.name) for src, _data in presets
+        if (vdir / expected.get(src, src.name)).stat().st_size == 0
+    ]
+    if empty:
+        raise VariantStagingError(
+            f"{len(empty)} staged variant file(s) are empty: "
+            f"{', '.join(sorted(empty)[:5])}")
+
+
 def copy_variants_to_mod_dir(
     presets: list[tuple[Path, dict]],
     mod_dir: Path,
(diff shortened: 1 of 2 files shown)


faisalkindi:
Reviewed independently in an isolated worktree, by injecting failures rather than reading the diff — the whole value of this PR is in the failure path.

**The core is right and I have direct evidence for it.** I monkeypatched copy_variants_to_mod_dir to write the first variant then raise OSError(28, no space left), driving the real import_from_zip flow against a real file-backed Database, not a stub. Afterwards: the previously installed mod still present, still enabled=1, mod_deltas rows intact, files byte-identical, no .import-* or .replaced-* residue. Confirmed through a fresh sqlite3.connect so it's committed on-disk state rather than a dirty handle.

The DB rollback is real, which I checked rather than assumed. database.py:106 connects with the default isolation_level, so writes open an implicit transaction and rollback genuinely reverts them; the single commit sits after the filesystem swap. I forced two failures after DELETE FROM mod_deltas had already run — a failing swap rename, and a failing commit — and mod_deltas went 3 back to 3 both times with files restored byte-identical.

Cross-device rename is a non-issue by construction: staging is created inside mods_dir, so both renames are same-directory. The partial-swap window is two adjacent metadata-only renames, versus the original bug's window spanning an entire multi-file copy.

**I pushed one fix to this branch before merging, because it re-creates the exact data loss the PR exists to eliminate.**

The restoring os.replace(backup, mod_dir) runs inside suppress(Exception), so when THAT rename fails the following backup = None never executes. The finally-block then deletes the backup while mod_dir is still missing, since restoring it is precisely what just failed. That is the user's only remaining copy. The comment saying the branch is only reached on the success path isn't correct.

Reproduced before fixing, with both renames failing:

    raised    : PermissionError
    fs before : merged.json, variants/alpha.json, variants/beta.json
    fs after  : None          <- mod directory gone
    db intact : True          <- DB still points at a directory that no longer exists
    leftovers : none          <- backup deleted, nothing left to recover from

A BaseException (KeyboardInterrupt, SystemExit) after the backup rename reaches the same place, skipping except Exception while finally still runs.

The fix deletes the backup only once mod_dir is confirmed back in place, and otherwise leaves it and logs the path to move it back. I also stopped the except-block asserting it restored the previous mod directory when the restore failed — that line is what someone reads while trying to recover, so it shouldn't claim a recovery that didn't happen. New test drives both renames to failure; it fails on the previous code with the directory gone and the backup deleted, and passes now.

**One thing this PR does not do, so you know:** acceptance criterion 2 from #314 is untouched. The failure message is still the misleading generic text telling the user to drop the zip on the main window — it never says the update failed or that the existing install was kept, and it advises retrying the thing that just failed. It comes from import_handler.py:3510-3517 via the except at :4611, and this PR only changes variant_handler.py. Worse, the three GUI drag-drop sites (fluent_window.py:4404, :4648, :4930) catch into a bare logger.error and show nothing at all. Both pre-existing; I'll file them separately rather than hold this.

Two smaller things worth a later look: the stale-backup rmtree at :659 will destroy a previous crash's leftover if a PID is recycled, and nothing sweeps orphaned .import-*/.replaced-* under CDMods/mods (temp_workspace sweep_stale only scans the system temp dir), so a hard kill leaks them silently.

Tests: the PR's rollback suite 8 passed including my new case; variant/reimport sweep 116 passed, 1 skipped; full suite 2496 passed, 43 skipped, with only the known pre-existing py-launcher failure. New test file is ruff-clean; variant_handler.py has 12 findings identical to master.

Credit: AgentKush.

PR #334, opened 2026-07-30
Add an apply-level gap scanner (coverage_scan reports a false all-clear)

AgentKush:
`scripts/coverage_scan.py` asks whether `validate_intents` **claims** it can apply an intent. That is a different question from whether bytes come out — and the difference is where this project's real bugs have lived:

- equipslotinfo `etl_hashes` validated clean and wrote nothing
- iteminfo `match` selected zero records
- skill `use_resource_stat_list[N].d` applied 0 bytes
- characterinfo `call_mercenary_spawn_duration` was dropped while its sibling applied, so "No CD Mount" mods half-worked (#333)

Every one of those passed the validate-level scan. I built a fresh Nexus corpus — 192 archives, 498 JSONs, 52 Format 3 mods — and ran it:

```
No uncovered Format-3 intents found across the scanned mods.
```

The apply path was meanwhile dropping **1466 intents**.

## What this does instead

Drives the real apply path — `expand_format3_into_aggregated`, every whole-table writer included — and reports what the engine **itself** logged as skipped, refused or unresolved. Nothing here re-derives that judgement: the writers already say which field on which record they couldn't place, and that log is the ground truth.

Output on the corpus:

```
Scanned 52 Format-3 mod(s) against committed fixtures.

NOT SCANNED -- no vanilla bytes available for these targets:
  stringinfo.pabgb
  (these are unmeasured, not proven clean)

Dropped at apply time (the engine's own reason):
  x1452  iteminfo(1.13): nested path 'sharpness_data.stat_data.stat_list_static' on key=N skipped (unresolved)
  x1     characterinfo: could not locate field 'character_weight' for entry 'Kliff_Clone' ...
  x1     characterinfo: could not locate field 'f36' for entry 'Kliff_Clone' ...
```

## Design notes

**Runs in CI with no game install.** Vanilla bytes come from the committed `tests/fixtures/vanilla*` snapshots — 8 of the 9 tables real mods target have one. `--game-dir` scans a real install instead, which covers the ninth.

**Same CLI shape and exit-code contract as `coverage_scan.py`** (exit = number of distinct drop classes), so CI can gate on it identically.

**A target whose vanilla bytes can't be sourced is reported NOT SCANNED, never counted as clean.** A scanner that goes quiet about what it couldn't read is the exact failure mode it exists to catch, so that property has its own test.

**Per-record noise collapses into one class.** One field refused across 1400 records is one gap, not 1400 — otherwise the report is unreadable and the exit code is a record count.

**Mod detection goes through `parse_format3_mod_targets`**, like `coverage_scan`. My first cut sniffed for a `format: 3` key and found 4 mods out of 498 instead of 52; DMM ships wrapper shapes that carry no such key.

## Verification

Removed the #333 fix and re-ran — the scanner reports `x14 call_mercenary_spawn_duration is not supported`. Put it back — the drop disappears. So it detects a real regression rather than printing boilerplate.

12 tests cover the honesty properties: drops reported, unscanned surfaced, exit code equals distinct drop classes, a crashing mod becomes a reported gap rather than a lost scan, INFO chatter never becomes a gap, and an unformattable log record doesn't abort a corpus scan.

Full fast suite 2633 passed, 42 skipped. Ruff clean on both new files.

## Scope

Read-only and additive — no engine behaviour changes, nothing wired into CI here. Wiring it into the workflow is your call; it's a one-line addition next to the existing `coverage_scan` step if you want it gating.

One caveat worth stating: the 1452 `sharpness_data` drops are a **decode** gap, not a wiring one. All 6508 records read `sharpness_data = None`, so the decoder never populates it for anything — CDUMM is refusing correctly rather than writing to a guess. Closing that is RE on the equipment record, which I haven't attempted here.


AgentKush:
Pushed two follow-ups after re-checking my own claims.

**1. The scanner was silent about the mods it doesn't cover.** It's Format-3 only, and said so nowhere in its output. Measured on the corpus:

| | files | share |
|---|---|---|
| v2 byte-patch | **226** | 45.4% |
| other JSON (not a mod) | 214 | 43.0% |
| **Format 3** | **52** | 10.4% |

So `No intents dropped` was covering about a tenth of the mod files by count while reading like a clean bill of health for the whole population. v2 applies through `aggregate_json_mods_into_synthetic_patches` — a path this doesn't touch. The summary now prints the count and states the scope:

```
Scanned 52 Format-3 mod(s) against committed fixtures.
Skipped 226 v2 byte-patch mod(s): this scanner covers Format 3 only, and v2 applies through a different path.
```

Same reasoning as the existing NOT SCANNED line — a scanner that goes quiet about what it skipped is the failure mode it exists to catch.

I found this by chasing a discrepancy rather than by design: 9 corpus files mentioned `dropsetinfo`/`gimmickinfo` yet neither table appeared in the Format-3 list. Checked all 9 individually — 9 NOT Format 3, 0 Format 3, and none using `op=scale` or a `[*]` wildcard. So the detector was right; the silence wasn't.

`count_v2_mods` matches both shapes that appear on Nexus: an explicit `"format": 2`, and mods carrying `patches` with no `intents` that omit the key. Cross-checked against a separately written census — both report 226.

**2. The lint failure earlier was mine, and worth recording.** `EXE001 Shebang is present but file is not executable` is a Unix-permission rule, so local ruff on Windows cannot observe it — it passed here and failed the Linux runner. `coverage_scan.py` carries the same shebang at mode 100755; `apply_scan.py` had landed 100644. Fixed with `git update-index --chmod=+x`, and lint is green now.

15 tests, ruff clean on both files.

AgentKush:
Third follow-up: added the `vanilla115` stringinfo fixture, which closes the scanner's last measurement blind spot.

`stringinfo` was the only Format-3-targeted table with no committed snapshot, so the scanner reported it NOT SCANNED and its 174 real intents went unmeasured. With the fixture it covers **all 9 tables real Format 3 mods target** and the NOT SCANNED section disappears entirely:

```
Scanned 52 Format-3 mod(s) against committed fixtures.
Skipped 226 v2 byte-patch mod(s): this scanner covers Format 3 only, and v2 applies through a different path.

Dropped at apply time (the engine's own reason):
  x1452  iteminfo(1.13): nested path 'sharpness_data.stat_data.stat_list_static' ... (unresolved)
  x14    characterinfo: field 'call_mercenary_spawn_duration' is not supported, skipping
  x1     characterinfo: could not locate field 'character_weight' for entry 'Kliff_Clone' ...
  x1     characterinfo: could not locate field 'f36' for entry 'Kliff_Clone' ...
```

**stringinfo produces zero drops** — all 174 intents apply. No new gap found, but it's measured now rather than assumed, which is the point.

Two notes on reading that output, so nothing is over-read:

- The `spawn_duration` line is expected here and is **not** a regression. This branch is cut from master, and the fix for it is #333 — still an open PR. On a tree with #333 applied, that line is absent. That round-trip is also how I verified the scanner detects a real regression rather than printing boilerplate.
- The two `Kliff_Clone` lines are #332's guard doing its job, not a defect.

On the binary: 553 KB compressed for both files, comparable to the committed `iteminfo.pabgb.zlib` at 549 KB, same zlib format and naming so `fixture_loaders` picks it up with no code change. Before committing I verified the zlib round-trip is exact and that `load_vanilla115` returns bytes byte-identical to the live 1.15 install for both files (1,856,630 and 248,514 bytes).

**It's a separable commit** — if you'd rather not carry another binary, drop just that one; the scanner works without it and simply keeps reporting stringinfo as unmeasured.

Full fast suite 2633 passed, 42 skipped with the fixture in place.

AgentKush:
**Correction — I got the sharpness_data diagnosis wrong above, and the conclusion changes.**

In the PR body I wrote that the 1452 `sharpness_data` drops are a *decode* gap, on the grounds that "all 6508 records read `sharpness_data = None`, so the decoder never populates it". **That measurement was invalid.** I called `parse_iteminfo_from_bytes` without a `fields=` argument, so it used a default field list rather than the one the apply path selects via `detect_iteminfo_layout`. Under that wrong list *both* `sharpness_data` and `enchant_data_list` came back absent — and `enchant_data_list` demonstrably works (14,899 intents apply), which is what exposed my error.

Re-measured the way the apply path actually decodes (`fields = detect_iteminfo_layout(body, offsets)`, 119 fields):

```
sharpness_data in layout:    True
enchant_data_list in layout: True
records decoded: 6508
  sharpness_data present as a dict on:  6508 / 6508
  keys: craft_tool_info, max_sharpness, p_prefix, shape,
        stat_list, tail, w_trailing, w_unk_a
```

So `sharpness_data` **is** decoded on every record. There is no `stat_data` sub-key — but there is `stat_list`.

**It looks like a naming difference, not a missing decode.** The mod writes DMM's nested path; CDUMM decodes a flatter shape:

| | |
|---|---|
| mod writes | `sharpness_data.stat_data.stat_list_static = [{"stat": 1000003, "change_mb": 8000}]` |
| CDUMM decodes | `sharpness_data.stat_list = [{"stat": 1000003, "change_mb": 1000}]` |

Element keys are identical — `('change_mb', 'stat')` — and that sample is the *same record the mod targets*: key 14510 `Marni_Devotee_PlateArmor_Helm`, same stat id 1000003, 1000 → 8000. Across the 1269 targeted records **every one has a non-empty `stat_list`** (945 with one entry, 324 with two), all shape `W`. And `_read_ItemInfoSharpnessData` reads a single flat `u32 stat_count + N*12 stats`, so `stat_list` is the only stat list under `sharpness_data` — there is nothing else the DMM path could mean.

That is the same shape of problem as the characterinfo `character_weight` / `default_action_action_index` rename, not a blocked RE task.

**What I am not yet claiming:** that a path alias is correct to ship. Matching element shapes and a matching sample are strong, but I haven't verified a write round-trips byte-exact, and I'm not going to assert it before I have. I'm working that now and will report either way.

Nothing in this PR depends on the retracted claim — the scanner reported the drops correctly; my interpretation of *why* was wrong.

AgentKush:
## Composition check: all five PRs merged, measured on one corpus

I merged #320, #332, #333, #334 and #335 into a scratch branch off `master`
(`b5957d3`) and measured the result, rather than assuming they compose.

**Merge:** all five applied with **zero conflicts**, including #332 and #333
which both edit `characterinfo_writer.py`.

**Suites on the merged tree** (local, Windows, Python 3.13):

```
fast   2703 passed,  42 skipped,   39 deselected   (5:08)
slow     38 passed,   1 skipped, 2745 deselected   (9:58)
```

Zero failures.

### Apply-level drops, before and after

Both runs use the same scanner and the **same 138-mod Format 3 corpus**, so
they are directly comparable. (v2 mods are skipped by the scanner — 268 of
them — because they apply through a different path.)

| | `master` b5957d3 | all five merged |
|---|---|---|
| drop events | **1476** | **10** |
| distinct classes | 10 | 8 |

What each PR accounts for:

| PR | drop events it removes |
|---|---|
| #335 iteminfo `sharpness_data` alias | 1452 |
| #333 characterinfo `spawn_duration` | 14 |
| #320 statusgroupinfo `status_info_list` | 1 (`Global_Critical_Rate`) |
| #332 partial-record guard | **+1** (adds its own refusal line) |
| #334 the scanner | 0 — it is the measuring instrument |

#332 raising the count by one is the intended behaviour, not a regression:
it converts a *silent* half-written `Kliff_Clone` record into a loud,
all-or-nothing refusal, and prints why.

### What the remaining 10 events are

Three separate things, and only one of them is an unknown:

1. **3 events — #332's deliberate refusal.** One record (`Kliff_Clone`) in
   one mod. `character_weight` and `f36` don't resolve on it, so the record
   is left alone instead of being written with 4 of its 6 fields. Working
   as designed.

2. **1 event — `elementalmaterialinfo`.** The gap we already know is open on
   purpose. The scanner also lists it under `NOT SCANNED` because there's no
   committed fixture for that table, so it is reported twice by two different
   honesty mechanisms — which is the right outcome.

3. **6 events — `storeinfo.pabgb`, and this one I had not reported before.**
   Two Black Market mods (`mod3107_12503`, `mod3107_12504`) each `set` a whole
   new element into `stock_data_list`:

   ```json
   {"entry": "Store_Dem_BlackMarket", "key": 2009,
    "field": "stock_data_list", "op": "set",
    "new": [{"raw_a": 1000000, "raw_b": 1000000, "raw_c": 999,
             "value": {"disc": 0, "lookup_a": 0, ...}, ...}]}
   ```

   The writer refuses because the nested `value.raw_a` has no mapped binary
   position. It refuses rather than corrupting the table, which is correct —
   but it means "add an item to a store" doesn't work yet. That is the next
   real gap, and it is a *new-nested-record* problem, not a scalar-field one.

### One correction, and one design question

**Correction.** An earlier comment on this PR showed scan output taken from a
smaller 52-mod slice of the corpus, not the full 138. The `stringinfo produces
zero drops` claim in it still holds — there are no `stringinfo` lines in either
run above — but the drop list in that comment was not the whole picture. The
table above is the full measurement. I'd rather flag that than leave two
different numbers in the thread.

**Design question for you, not something I'll change unilaterally.**
`apply_scan` exits with the number of distinct drop classes, so this branch
exits `8` even though 3 of those events are a deliberate refusal and 1 is a
known-open table. If this is ever wired into CI as "exit 0 or fail", a correct
refusal is currently indistinguishable from a real gap. Options I see:
classify refusals separately from unresolved-drops, or take a baseline file
and fail only on new classes. Happy to implement whichever you prefer.

Reproduce with:

```
python scripts/apply_scan.py <dir-of-format3-mod-jsons>
```

PR #336, opened 2026-08-02
Add the CD 1.16 iteminfo layout (cd116) - 0 to 6,581/6,581 byte-exact

AgentKush:
CD **1.16.00** (Steam buildid 24504786, exe 1.0.0.2258) changed the iteminfo
record. Every existing layout round-trips **0 of 6,581** records, so
`detect_iteminfo_layout` returns `None`, every record falls to the per-record
opaque path, and all field-level item mods are refused. Byte-safe, but nothing
decodes.

Measured against the live 1.16 table, whole-table, each step cumulative:

| layout | byte-exact |
|---|---|
| `cd113_enchant` (unchanged) | **0 / 6,581** |
| + drop `inventory_info` | 5,579 / 6,581 (84.8 %) |
| + drop `repair_data_list` | 6,190 / 6,581 (94.1 %) |
| + drop the sharpness `pre_unk_109` byte | **6,581 / 6,581 (100 %)** |

### The three changes

**1. `inventory_info` (u16) is gone.** Absent from ItemInfo's field list in the
1.16 binary, and from every other type in it.

**2. `repair_data_list` is no longer a carray.** The u32 the old codec read as
its element count is `0` on every passing record and `0xFFFFFFFF` or `604800`
(= 7 days in seconds) on the rest — field values, not a length. That is also
why the old codec *looked* fine: a zero reads as "empty list". Its bytes are
carried as `_tail_slack`, so the round trip stays byte-exact.

**3. `sharpness_data`: 1.16 removed the CD 1.09 conditional pre byte** (the u8
taken when `default_sub_item.type_id < 14`). Proof on key 14510
(`Marni_Devotee_PlateArmor_Helm`): without it the W-header reads
`max_sharpness=40, stat_count=1, stat=1000003, change_mb=1000` — matching the
ground truth established independently on CD 1.13. With it, `max_sharpness`
reads 16640.

All three are the same failure mode: **a field whose bytes are normally zero,
which lets a stale codec look correct until it meets real data.** Worth checking
for first on the next version break.

### One dispatch change worth reviewing

`_read_item`'s sharpness branch now tests
`spec[2] is _read_ItemInfoSharpnessData` instead of
`name == "sharpness_data"` (two call sites: the reader and the skip/measure
walker). Keying on the *reader* rather than the *name* lets `cd116` supply a
different codec under the same field name. Matching on the name would have
forced a renamed field, and every mod targeting `sharpness_data` would break.

### Verification

Through the real detector, no monkeypatching:

| table | detector picked | round-trip | whole-table |
|---|---|---|---|
| 1.10 fixture | `default` | 6,419 / 6,483 | byte-exact |
| 1.13 fixture | `cd113_enchant` | 6,508 / 6,508 | byte-exact |
| 1.16 live | `cd116` | 6,581 / 6,581 | byte-exact |

No version check anywhere — detection scores layouts by round-trip, so a future
build matching none still falls back to opaque rather than corrupting.

Fast suite: **2566 passed, 42 skipped**.

Also re-verified against the 2026-08-02 hotfix (buildid 24514474, exe
1.0.0.2264): all 270 data tables byte-identical to 1.16, `cd116` still
6,581 / 6,581.

### Notes

* `cd116` is built **without** the 1.13 prefab tail on purpose — 1.16 moved that
  region too, and carrying it opaque scores better (6,581 vs 6,567). Deriving it
  is follow-up work.
* No conflicts with #320, #332, #333, #334 or #335 (test-merged; none touch
  `iteminfo_native_parser.py`). #335 aliases a nested path on the parsed dict,
  which `cd116` leaves unchanged, so the two are independent.


AgentKush:
**Re-verified against a newer game build.**

Crimson Desert patched again mid-review — **buildid 24534261** (2026-08-03); this PR was measured on 24504786 / hotfix 24514474.

Re-ran detection and the whole-table round trip on the new install:

```
1.13 fixture           picks cd113_enchant   6508 records   whole-table byte-exact: True
LIVE build 24534261    picks cd116           6581 records   whole-table byte-exact: True
```

Unchanged. The `cd116` layout is still selected by the real detector and still reproduces the entire table byte-for-byte.

AgentKush:
**The opaque tail is now derived — `prefab_data_list` is back on 1.16.**

The original cut of this PR stopped `cd116` at `max_endurance` and carried everything after it opaque: **713,449 bytes, 11.6% of the live table, on every one of the 6,581 records.** I called deriving it "follow-up work". It's done, and it's in this branch now.

### What 1.16 actually did

It didn't change the repair/prefab region — it **wrapped** it. Diffing the same item key between the 1.13 fixture and the live 1.16 table shows the 1.13 region appearing *verbatim* inside 1.16's undecoded remainder, at offset 10:

```
1.16 tail = [10 bytes] + [the 1.13 repair + prefab region] + [18 bytes]
```

replacing 1.13's 2-byte `record_end_a`/`record_end_b`.

Checked on the items where it *can* be checked — the **236 whose `equip_slot_list` is non-empty**. Three empty carrays are the same twelve zero bytes in any order, so items with empty lists can't distinguish a reorder from no reorder. All 236 that can, match. The element layout is unchanged.

### Why 10 and 18 aren't fitted

Sweeping `pre` 8..12 against `suf` 16..20:

| | result |
|---|---|
| `pre=10, suf=18` | 14 opaque, 0 slack, **12,262 prefab elements** |
| every other cell | 6,581 opaque, 0 elements |

One cell works and its neighbours fail *completely*. A fitted parameter doesn't behave that way.

Independent corroboration from the element count: 1.13 has **12,155** across 6,508 records; 1.16 gives **12,262** across 6,581. +107 for 73 new items is the right magnitude — a self-consistent misread has no reason to land there.

### Result, through the real detector

| build | picks | opaque | slack | prefab | byte-exact |
|---|---|---|---|---|---|
| 1.13 fixture | `cd113_enchant` | 0 | 0 | 12,155 | ✓ |
| 1.10 fixture | `default` | 64 | 0 | 5,854 | ✓ |
| **1.16 live** | **`cd116`** | **14** | **0** | **12,262** | ✓ |

`_tail_slack` 713,449 → **0**. `prefab_data_list` 0 → **12,262**.

That field matters: community-3.5.0.14 shipped a writer for it (Equip Everything / AXIOM QoL tribe-gender equip restrictions), so those mods were being refused on 1.16.

### Notes

* The 14 stragglers all fail identically (`carray count N exceeds 46 remaining bytes`) and fall to the opaque-record path — still byte-exact, just uneditable.
* Verified version-specific: this layout on a 1.13 table yields **6,508 opaque and zero elements**, so it can never be selected there.
* The 28 wrapper bytes are two *named* opaque runs, not decoded fields — what they hold is still unknown. Naming them is precisely what lets the fields between them parse.
* `_with_cd116_tail` takes a `source` argument on purpose: `_ITEM_FIELDS_CD113` has `repair_data_list` but **not** `prefab_data_list` (only the enchant variant has both), so sourcing from the wrong list silently drops the prefab codec. That cost me a debug cycle; the parameter makes it impossible to get wrong.
* Zero new lint findings (23 before, 23 after, all pre-existing).

PR #338, opened 2026-08-02
storeinfo: locate the stock list instead of computing its offset (fixes 82 stores read as empty, + the CD 1.16 shape)

AgentKush:
Three problems with one root cause: **the stock list's `u32` count does not sit at a fixed offset into the entry payload.** Some entries carry an extra block before it, so a single game build holds counts at four different offsets (44, 92, 96, 112 on CD 1.13).

### 1. A shipped bug on current stable

`count_payload_offset` is one constant. For **82 of the 1.13 table's entries** CDUMM therefore read a *different* `u32` — usually `0` — and reported "this store is empty".

That is a **successful parse of the wrong bytes**: no exception, no warning, no refusal. **1,782 stock records invisible across 82 stores.**

The write path is worse. A Format 3 mod targeting one of those stores would have spliced a new list in at `payload+44`, i.e. into the middle of live fields, on a table whose only integrity check is the game crashing on store open.

### 2. CD 1.16 broke the record shape again

A `u32` after `raw_c` — `_lowPriceThresholdCount` in the binary's own field names. Under the 1.13 shape the live 1.16 table decodes **zero** entries, so every store mod silently stopped applying. Third time this record has moved (1.11 added `is_restore_item`, 1.13 added `order_index_113`).

### 3. `_orderCountDataList` was refused outright

"element layout not decoded" — costing 15 entries of the 1.13 table and 10 of 1.16.

## Finding the list instead of computing it

`StockData._storeInfo` is a `u16` back-reference to the owning store, and it is the first field of the first record. So the list is anchored on the entry's own key. Scanning for that — plus a plausible count and a byte-exact round-trip of the whole span — pins the list with **no offset constant at all**.

Measured over both real tables, **exactly one offset per entry** satisfies all four conditions. Never two.

The anchor is not just a tiebreak. On 10 entries a span elsewhere *does* round-trip as a valid 1-record list, and the anchor is what rejects it — its `lookup_a` is `0`, not the store key.

## What "provably empty" buys

An anchor can only find a *non-empty* list, so entries with no stock need a separate answer — and "we couldn't find one" is exactly the silent gap that caused problem 1.

It turns out every such entry is **too small to hold one record at any offset**: 113 bytes at most against a 182-byte minimum on 1.13, 114 against 186 on 1.16. They are provably empty rather than unread. `StoreListNotFound.provably_empty` carries that distinction, and adding stock to such a store is refused with a message that says why instead of guessing an offset.

That is the difference between *"293 of 293 entries decoded or provably empty"* and *"263 decoded, 30 we didn't get to"*.

## Order element size, derived not assumed

`StockOrderCountData` is 12 bytes, found by **exact tiling**: walk the whole list assuming a fixed element size and require every following record to land on a valid boundary. Exactly one size in 1..128 works, and it works for every affected entry — unique-or-nothing, not best-fit.

The interior is carried verbatim. The schema names two fields inside it, but there is no ground truth for either, so decoding them would be a guess.

## Results

Byte-exact through the shipped detector, no monkeypatching:

| table | before | after |
|---|---|---|
| CD 1.13 fixture | 166 entries / 3,661 records | **263 entries / 5,443 records** + 30 provably empty |
| CD 1.16 live | 0 entries / 0 records | **397 entries / 6,376 records** + 35 provably empty |

**293/293** and **432/432** entries accounted for — located byte-exact, or provably empty. There is no third category.

## The CD 1.16 table is committed

`tests/fixtures/vanilla116/` (64 KB zlib, same order as the existing `vanilla113` fixture). It had to be: this table's shape drifted two whole game versions with CI green, because the only 1.16-era table lived on whichever machine had the game installed.

Both fixtures are now pinned by the same assertions, including one that would have caught all three of these bugs — *every entry is either located byte-exact or provably empty*.

## Notes for review

* `detect_storeinfo_layout` no longer raises on an all-empty table. It separates "too small to hold a record under any layout" (valid, degenerate, unknowable by construction) from "big enough, but nothing decodes" (unknown shape, still refuses).
* `StoreLayout.count_payload_offset` survives as documentation and for building synthetic entries; nothing reads it to find a list any more.
* `test_writer_name_lookup.build_store_table` now sets `lookup_a` to the entry key, because the real format does. The synthetic tables were the only place that didn't.
* The scan is cheap — count plausibility and the anchor are 6 bytes, checked before any parse. Detection over both whole tables runs in about a second.


AgentKush:
**Re-verified against a newer game build.**

Crimson Desert patched again mid-review — PAZ archives rebuilt 2026-08-03, **buildid 24534261** (this PR was measured on 24504786 / hotfix 24514474). Re-ran the same checks against the new install:

- `storeinfo.pabgb` is the **same size and same header** but a **different SHA** — a values-only patch, structure untouched.
- Detector still picks **CD 1.16**.
- **432/432 entries accounted for** — 397 located byte-exact + 35 provably empty, 6,376 stock records, 1,063 order-count elements. Identical to the numbers in the description.

The committed `vanilla116` fixture is therefore one build behind on *content*, but structurally identical and every assertion still passes, so I have left it alone rather than churn the diff.

PR #339, opened 2026-08-02
Choose the semantic walker's field order against the table (iteminfo 10 -> 109 fields on CD 1.16)

AgentKush:
There are **two independent iteminfo readers**, and fixing one does nothing for the other:

* `iteminfo_native_parser` — hand-written, and the one given a CD 1.16 layout in #336;
* the **generic walker** driven by `_ordered_fields`, which is what the Format 3 generic path uses: `match`, nested targets, the field grid.

CD 1.16 removed two ItemInfo fields. A removed field is the worst drift there is — the walker keeps reading, takes the *next* field's bytes as the missing one, and desyncs everything after it. Measured with `schema_verify.decode_score` on the live table:

| build | median fields decoded per record |
|---|---|
| CD 1.13 fixture | 110 of 113 |
| **CD 1.16 live** | **10 of 113** |

That is the "11-field iteminfo grid" wall returning one game version later. Nothing went red, because the only 1.16-era table lived on whichever machine had the game installed.

## What changed

The order is now **scored against the table in front of us**, the same way iteminfo's record layout and storeinfo's already are. `ORDER_VARIANTS` declares the fields a build doesn't have; the declared order is always a candidate and only loses to a variant that decodes **strictly better on both axes**, so a tie keeps the order that was actually reverse-engineered against real bytes.

| table | picks | result |
|---|---|---|
| iteminfo 1.10 fixture | `base` | 64/113 |
| iteminfo 1.13 fixture | `base` | 110/113 |
| iteminfo 1.16 live | `cd116` | **109/111** |

The two removals are the same ones the native parser needed to go from 0 to 6,581 byte-exact records — two independent decoders arriving at the same 1.16 change, which is the corroboration that makes this more than a curve-fit.

## Blast radius, deliberately small

`schema_for_table` returns **the very same object** `get_schema` returns — not an equal one, the same one — for every table that declares no variant, which is all of them except ItemInfo. That's asserted per-table:

```python
assert sv.schema_for_table(table, b"", b"") is get_schema(table)
```

Scoring walks every record of a 6 MB table (0.4–0.8 s), so it's cached per table body; a test pins that it happens once, not once per intent.

## Adding a future build

One line in `ORDER_VARIANTS`. Getting it wrong costs nothing — a variant that doesn't decode better is never selected. The tests include that negative directly: the `cd116` variant scores *worse* on a 1.13 table, so it cannot be picked there, and a variant that merely ties is rejected.

## Notes

* Independent of #336 and #338 — different file, different reader. #336 fixes what writes iteminfo records; this fixes what *reads fields out of them* for the generic path.
* `schema_verify` already imports `format3_apply`, so the new call is a function-local import at the single `get_schema` call site.
* This was found by sweeping every table with a verified field order against live 1.16 and comparing to the committed fixture. ItemInfo was the only one that decoded worse — `CharacterInfo`, `FieldInfo` and `VehicleInfo` are unchanged at 100%, and `StageInfo`/`RegionInfo`/`WantedInfo` are shallow on *both* builds, i.e. never fully modelled rather than newly broken.

### DIFF
--- a/src/cdumm/engine/format3_apply.py
+++ b/src/cdumm/engine/format3_apply.py
@@ -2103,15 +2103,21 @@ def _intents_to_v2_changes(
             "PABGH key_size=%d", target, key_size)
         return []
 
-    # Pick the field order that fits THIS table, not the one that fitted
-    # the build we last reverse-engineered. A patch that removes a field
-    # desyncs the walker from that field onward -- CD 1.16 dropped two
-    # ItemInfo fields and the walk fell from 110 of 113 fields to 10,
-    # which is the iteminfo grid going blank. Falls back to get_schema
-    # for every table with no variants declared. Imported here because
-    # schema_verify imports this module.
-    from cdumm.engine.schema_verify import schema_for_table
-    schema = schema_for_table(table_name, vanilla_body, vanilla_header)
+    # Order selection deliberately does NOT happen here.
+    #
+    # It was wired in at this line first, and that was the wrong place
+    # twice over. It is unreachable for the one table that has a variant:
+    # `iteminfo_force_batch` below routes every iteminfo intent to the
+    # native writer before `_resolve_write_pos`, the only consumer of
+    # `field_specs`. And selecting an order costs a decode of the whole
+    # table, which this function runs once per apply -- against a body
+    # that a previous intent has just modified, so the per-body cache
+    # misses every time and the cost lands on every mod applied.
+    #
+    # The symptom being fixed is a stale field order in the Game Data
+    # grid, so selection lives on that path (`parse_records_display`),
+    # where it runs once per table body and is actually reachable.
+    schema = get_schema(table_name)
     field_specs = {f.name: f for f in schema.fields}
     fs_entries = load_field_schema(table_name)
 
--- a/src/cdumm/engine/format3_apply.py
+++ b/src/cdumm/engine/format3_apply.py
@@ -2103,7 +2103,15 @@ def _intents_to_v2_changes(
             "PABGH key_size=%d", target, key_size)
         return []
 
-    schema = get_schema(table_name)
+    # Pick the field order that fits THIS table, not the one that fitted
+    # the build we last reverse-engineered. A patch that removes a field
+    # desyncs the walker from that field onward -- CD 1.16 dropped two
+    # ItemInfo fields and the walk fell from 110 of 113 fields to 10,
+    # which is the iteminfo grid going blank. Falls back to get_schema
+    # for every table with no variants declared. Imported here because
+    # schema_verify imports this module.
+    from cdumm.engine.schema_verify import schema_for_table
+    schema = schema_for_table(table_name, vanilla_body, vanilla_header)
     field_specs = {f.name: f for f in schema.fields}
     fs_entries = load_field_schema(table_name)
 
--- a/src/cdumm/engine/schema_verify.py
+++ b/src/cdumm/engine/schema_verify.py
@@ -260,11 +260,25 @@ class VerificationReport:
 #: variant that does not decode better is never selected.
 ORDER_VARIANTS: dict[str, tuple[tuple[str, tuple[str, ...]], ...]] = {
     "ItemInfo": (
-        # CD 1.16 removed _inventoryInfo (a u16) and stopped storing
-        # _repairDataList as a counted array. Confirmed independently by
-        # the native parser, which needed exactly the same two removals
-        # to go from 0 to 6,581 byte-exact records.
-        ("cd116", ("_inventoryInfo", "_repairDataList")),
+        # CD 1.16. What each removal is actually for:
+        #
+        # * _inventoryInfo (u16) is GONE -- absent from ItemInfo's field
+        #   list in the 1.16 binary.
+        # * _gimmickVisualPrefabDataList is GONE too -- likewise absent
+        #   from the 1.16 binary, which merged it into PrefabData.
+        # * _repairDataList and _prefabDataList still EXIST (they are
+        #   fields 110 and 111 of 115 in the 1.16 binary). They are
+        #   dropped here because 1.16 wrapped that whole region in 10
+        #   leading and 18 trailing bytes, and a field-NAME list cannot
+        #   express an opaque run -- the native parser handles it with
+        #   two named opaque fields, which this mechanism has no way to
+        #   name. Dropping them is what lets the walk continue past the
+        #   wrapper instead of desyncing on it.
+        #
+        # Do NOT read this list as "1.16 removed four fields". Two were
+        # removed; two are unreachable to a name-only order.
+        ("cd116", ("_inventoryInfo", "_repairDataList",
+                   "_prefabDataList", "_gimmickVisualPrefabDataList")),
     ),
 }
 
--- a/src/cdumm/engine/schema_verify.py
+++ b/src/cdumm/engine/schema_verify.py
@@ -49,6 +49,7 @@ import logging
 from dataclasses import dataclass, field
 from pathlib import Path
 from statistics import median
+from typing import Any
 
 from cdumm.engine.format3_apply import _consume_field_bytes, _payload_offset
 from cdumm.semantic import parser as parser_mod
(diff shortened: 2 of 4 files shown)


AgentKush:

Selection is unchanged:

```
iteminfo 1.10 fixture   picks base    64/113
iteminfo 1.13 fixture   picks base   110/113
iteminfo 1.16 LIVE      picks cd116  109/111
```

And the sweep across every table with a verified field order still names **ItemInfo as the only regression** — `CharacterInfo` 14/14, `FieldInfo` 19/19, `VehicleInfo` 20/20 unchanged; `StageInfo`/`RegionInfo`/`WantedInfo` shallow on both builds as before. Tables with no variant still return the identical `get_schema` object.

AgentKush:
**Improved: the variant now drops four fields, not two — 46.6% → 87.5% of records fully walked.**

Deriving the 1.16 tail (over in #336) showed my original justification here was half wrong, and fixing it made the variant better.

### What I had wrong

The comment said 1.16 "removed `_inventoryInfo` and stopped storing `_repairDataList` as a counted array". The first half holds. The second doesn't — **`_repairDataList` is still field 110 of 115 in the 1.16 binary**, and `_prefabDataList` is 111.

What 1.16 actually did to that region is **wrap** it: 10 leading and 18 trailing bytes around the *unchanged* 1.13 repair+prefab region. A field-**name** order has no way to express an opaque run, so the walker can't follow the wrapper and has to skip the region entirely.

### The four fields are two different categories

| field | why |
|---|---|
| `_inventoryInfo` | **gone** from the 1.16 binary |
| `_gimmickVisualPrefabDataList` | **gone** from the 1.16 binary (merged into PrefabData) |
| `_repairDataList` | **exists**; inside the wrapped region |
| `_prefabDataList` | **exists**; inside the wrapped region |

Please don't read the list as "the four fields 1.16 removed" — two were removed, two are unreachable to a name-only order. There's a new test pinning that intent, because trimming the list back to "just the removed ones" is a natural-looking edit that halves the walk depth.

### Measured

```
-_inventoryInfo -_repairDataList                     109/111   46.6% full
+ -_prefabDataList -_gimmickVisualPrefabDataList      109/109   87.5% full
```

Same median; the fraction of records consuming *every* field nearly doubles — and that fraction is exactly what `select_order`'s tie-break scores on, so the four-drop variant wins on merit rather than by my asserting it.

Version-specificity unchanged:

| build | base | four-drop | winner |
|---|---|---|---|
| 1.10 fixture | 64/113 | 8/109 | base |
| 1.13 fixture | 110/113 | 8/109 | base |
| **1.16 live** | 10/113 | **109/109** | variant |

Zero new lint findings; 12 tests pass.

faisalkindi:
Reviewed independently. **Holding this one: as written it never executes on the real path.** The derivation work behind it is good, so this is about wiring and about which metric decides, not about the analysis.

## It is unreachable, three separate ways

**1. The lookup key never matches.** `ORDER_VARIANTS` is keyed `"ItemInfo"` (`schema_verify.py:262`), and `select_order` does a bare `if table not in ORDER_VARIANTS` at `:297` with no case normalisation — unlike `get_schema`, which lowercases. Production passes `identify_table_from_path(target)`, which yields `"iteminfo"`:

```
ORDER_VARIANTS keys        : ['ItemInfo']
production passes          : 'iteminfo'
'iteminfo' in ORDER_VARIANTS : False
'ItemInfo' in ORDER_VARIANTS : True
```

Every new test calls `select_order("ItemInfo", …)` — the CamelCase spelling production never uses. That is why they pass green over code that cannot run.

**2. Even with the key fixed, the schema is unused for iteminfo.** `iteminfo_force_batch` (`format3_apply.py:2135`) routes every iteminfo intent to the native writer at `:2151`, ahead of `_resolve_write_pos` — the only consumer of `schema`/`field_specs`. Forcing the variant reachable and running a real primitive intent on live 1.16 gives 1 change and **0** calls to `_resolve_write_pos`.

**3. The grid it blames is a different function.** `parse_records_display` (`semantic/parser.py:657`, via `game_data_page.py:491`) calls `get_schema` directly and is untouched here. Same for the `match` path.

Consistent with all that, the measured regression is exactly zero — all 322 schema tables return the identical object, 15/15 fixtures identical, live 1.16 iteminfo identical. Zero values changed because nothing changes.

## The deeper reason I don't want to merge it as-is

It makes `decode_score` the arbiter of field order, and that metric is provably blind to the error class we just shipped a fix for.

Checked against the native parser as oracle on live 1.16, 6,567 records, 67 comparable scalar fields: 93.4% of values agree, but two fields agree on **zero** records:

```
_respawnTimeSeconds   0 / 5756 agree   walker reads 6876556153801630819
_maxEndurance         0 / 5756 agree   walker reads {0, 28515, 1}
                                       native reads 65535 on 6,445 records
```

That is the same 10-byte run I moved in #336, and the same `max_endurance` sentinel. A field-name order cannot express an opaque run — the PR's own comment concedes this — so the walker lands those two fields 10 bytes off while scoring median 109/109 with 87.5% full depth. `decode_score` consumes the same bytes either way and cannot see it.

**In fairness, this is not a regression the PR introduces.** I checked the declared order against native on the 1.13 fixture: `_maxEndurance` 0/5356 and `_respawnTimeSeconds` 0/5356 there too, with the same garbage values, while `_maxStackCount`/`_itemType`/`_isBlocked` are 6508/6508. The declared order was already wrong on those two. This PR neither breaks nor fixes them. My objection is narrower: it promotes a walk-depth metric to *the* decider of field order at the moment a value oracle finally exists on master, and doesn't consult it.

## The derivation itself is sound

Worth saying so. On live 1.16 an exhaustive probe of all 52,360 four-subsets of the 35-field neighbourhood makes cd116 uniquely best (median 109, reach 0.876, against a runner-up at 108/0.481), a greedy search converges elsewhere and never passes median 28, and it correctly loses on older builds. That is real work.

One latent issue in the selection rule: `_select_order_uncached` (`schema_verify.py:318-325`) uses strict `>`, so among variants tied at the top the first declared wins — no uniqueness requirement and no refusal on ambiguity. Safe today only because exactly one variant exists, and note 8,013 other 4-drop-sets also beat base on 1.16, so "strictly better than base" is a low bar. Same first-fit shape I just fixed in #340.

There is also no refusal path at all: base decoding at median 10 of 113 with no better variant proceeds silently, and a winning variant logs only at info level.

## What would make me merge it

1. Normalise the lookup case so production's `"iteminfo"` finds the variant.
2. Wire it where the visible symptom actually is — `parse_records_display` — rather than `format3_apply.py:2083`, which `iteminfo_force_batch` makes unreachable.
3. Gate selection on agreeing with the native parser's **values**, not on walk depth. The oracle exists now; #336 established it byte-exact.

Tests: the new file 12 passed; full suite 2,675 passed with only the known pre-existing py-launcher failure.

AgentKush:
You were right on all three, and the second one was worse than you said. Rebased onto master and reworked.

## 1. Unreachable — confirmed, and I hit it independently

Before touching this I was sweeping every 1.16 table and my own sweep scored **0 of 135** tables until I added a lowercase lookup. Same bug, found from the other side. `_variants_for` now normalises case the way `get_schema` does:

```
_variants_for('iteminfo')  -> True     (production's spelling)
_variants_for('ItemInfo')  -> True
_variants_for('storeinfo') -> None
```

The tests now use `"iteminfo"`, and one asserts both spellings select identically. Using the CamelCase spelling is what let the old file pass over dead code, so that is pinned rather than just fixed.

## 2. Wrong call site — and it was also a real slowdown

Moved to `parse_records_display`. On live 1.16, measured through `identify_table_from_path`:

| | median fields per record |
|---|---|
| `get_schema` (old wiring) | **10** |
| `schema_for_table` (new) | **111** |

I removed the `format3_apply` call site entirely rather than leave it. Beyond being unreachable as you showed, it made the apply pipeline much slower: `expand_format3` runs once per apply against a body a previous intent has just modified, so the per-body cache missed **every time**. My first full-suite run after wiring it took over an hour to reach 25% against 25:43 for the whole suite before. That is gone now — that path is byte-identical to master again.

## 3. Values, not walk depth

The oracle now decides. It reproduces your numbers exactly, including that this predates the PR:

```
1.13 fixture   base   401,739/428,195 (93.8%)   cd116   19,542/26,062 (75.0%)  -> base
1.16 fixture   base    23,116/ 26,268 (88.0%)   cd116  405,280/433,727 (93.4%) -> cd116

zero-agreement fields, 1.16 cd116 : _maxEndurance 0/5756, _respawnTimeSeconds 0/5756
zero-agreement fields, 1.13 base  : _maxEndurance 0/5356, _respawnTimeSeconds 0/5356
```

Comparison is sampled at 400 records spread evenly, because full comparison cost ~3s per table body and changed nothing: sampled and full agree to 0.1% and name the same zero-agreement fields. Record boundaries still come from the **full** entry list — taking them from the sampled list would hand each record a too-generous end and let a runaway count look survivable, which would make sampling change the answer rather than just the cost.

Cold `select_order` is 0.92s and cached after; grid open goes 1.09s -> 1.49s.

## On the points you raised but didn't require

- **Ties.** Now refuse and keep the declared order, rather than first-declared-wins. You called the same first-fit shape in #340 and it was the same mistake here.
- **No refusal path.** A table with no oracle now keeps its declared order instead of trusting a deeper walk on faith — that faith is exactly what produced the two zero-agreement fields.
- **The two bad fields.** Still wrong, still pre-existing, but no longer silent: selection logs `2 field(s) match the native parser on no record at all and should not be trusted in the grid: _maxEndurance, _respawnTimeSeconds`. A name-only order cannot express the opaque run, so fixing them means naming it, which belongs with the native layout rather than here.

## Verification

New tests 12/12; the iteminfo/entry-header/pabgh/format3-match related set 106 passed + 4 e2e; ruff clean on the added file with zero added debt on touched files (`schema_verify` 3 -> 3, `format3_apply` 25 -> 25, `parser` 5 -> 5, all pre-existing). Live-path check run against buildid 24545272.

Full local suite: **2790 passed, 43 skipped, 0 failed in 25:30** — matching the 25:43 baseline, so no lasting perf cost.

---

_(Reposting: the previous version of this comment had its backticks mangled by a shell-quoting mistake on my side, which broke the code fences and ate a couple of characters. Same content, correctly formatted.)_

PR #340, opened 2026-08-02
Derive the pabgh count width instead of looking it up by name (fixes 2 unreadable tables + a ciphertext fixture)

AgentKush:
A pabgh index is `[count][count × (key, u32 offset)]`, and the count prefix is **1, 2 or 4 bytes depending on the table**. Which one it was came from `UINT_COUNT_TABLES` — a hand-maintained set of table *names*.

That's the same failure mode as storeinfo's `count_payload_offset`: a per-table constant someone has to remember to add. `aimemoryinfo` needs a u32 count, isn't on the list, and so **its index doesn't parse at all** — which makes the entire table invisible to every writer, with nothing but a debug-level warning to say so.

## Deriving it

`len(header) == count_size + count × (key_size + 4)` has to hold with an integer key size. **That alone pins the layout for 130 of the 135 tables** in a current install.

The other five tile two ways, and all five have one or two entries. They're broken by a structural invariant: **entry offsets start at 0 and strictly increase**, because entries tile the body in index order.

I checked that invariant against all 130 unambiguous tables *before* using it — it holds for every one, so it's a property of the format rather than a rule fitted to five cases. It resolves `aimemoryinfo` outright, and the body confirms its answer independently (3/3 entries carry their own key at their own offset).

Where even that doesn't discriminate — four single-entry tables, where "offsets start at 0" is trivially true either way — `UINT_COUNT_TABLES` is consulted as a last-resort tiebreak, so those four behave exactly as before.

## Measured

Against a full CD 1.16 install and every committed fixture:

| | |
|---|---|
| live tables parsing identically to before | **133** |
| live tables that now parse and could not before | **2** — `aimemoryinfo` (key=4, 3 entries), `gamestartinfo` (key=1, 1 entry) |
| tables that **disagree** | **0** |
| committed fixtures parsing identically | 14 |
| degenerate headers (empty / short / garbage) | unchanged |

Shapes that turn up in the wild, all now covered: count 1/2/4 bytes; keys of 1, 2, 4, 8 — and a 12-byte composite (`aieventtableinfo`, which the old arithmetic already handled and which the new code must not break).

## A broken fixture, and the guard for it

While diffing, `tests/fixtures/vanilla113/inventory.pabgh` turned out to be **ciphertext** — 122 bytes of noise that parsed to nothing. Its body is fine, and the seven tests using it search the *body* by name rather than going through the index, so nothing ever went red.

Replaced with the correct header, and the replacement is provable rather than assumed: the vanilla113 and vanilla115 inventory **bodies are byte-identical**, and the 115 header anchors 20/20 entries against that shared body with names that decode (`Money`, `Character`, `PearlUser`, …).

Added the guard that would have caught it — **every committed `.pabgh` fixture must parse AND agree with its own body**, i.e. each entry's key must be the key stored at that entry's offset. Same anchor idea as the storeinfo work, turned on the fixtures themselves.

## Notes for review

* The riskiest change of the recent batch, since `parse_pabgh_index` is called everywhere — hence the full old-vs-new differential over 135 live tables plus all 14 fixtures rather than just a test or two.
* The `count == 0` path stays quiet: an empty table is valid, not an error.
* `gamestartinfo` now parses but has a 1-byte key, and `format3_apply` refuses `key_size` outside (2, 4), so it stays unaddressable by Format 3 — reading it correctly is still strictly better than failing to read it. Seven other live tables are in that category (1-byte keys) plus `characterappearanceindexinfo` at 8 bytes; widening that guard is separate work.
* Independent of #336, #338 and #339 — different file each time.


AgentKush:

The old-vs-new differential is unchanged on the new install:

```
133 live tables   parse identically to before
  2 live tables   now parse and could not before
                  (aimemoryinfo key=4 3 entries, gamestartinfo key=1 1 entry)
  0 tables        disagree
 14 fixtures      identical
    degenerate headers (empty / short / garbage) unchanged
```

PR #341, opened 2026-08-02
Read entry ids at key_size bytes for every width (unlocks 7 tables; two earlier diagnoses were wrong)

AgentKush:
Entry ids are `key_size` bytes wide. `_parse_entry_header` read them as u32 unless `key_size` was 1 or 2, which misread every table with a narrower or wider key.

### What was wrong

The failure mode is the one that keeps costing us tables: **the wrong read succeeds.** Nothing throws, nothing goes red, the table just quietly decodes to something else.

**1-byte keys.** `mercenaryinfo` has 1-byte keys, so a u32 read swallowed three bytes of the name and `Mercenary_Main` came back as `cenary`. That was recorded as evidence the table had a scrambled field order, and it was written off. At the right width it names 21 of 21 entries. Same for `gamestartinfo`, `socketinfo`, `relationinfo`, `terrainregionnaviinfo`.

**8- and 12-byte keys.** The first version of this PR left these on the u32 branch on purpose, arguing they carry no entry-name header so there was "nothing to check an answer against". That was wrong twice:

* There **was** an oracle all along — the PABGH index already gives the key for each offset, so the id read out of the body has to equal it.
* The header is not absent, only **empty**. It looked absent because the truncated id spilled into `name_len`, which tripped the `nlen > 500` guard and returned early.

| table | key | id==key before | after |
|---|---|---|---|
| `characterappearanceindexinfo` | 8 B | 0 / 8344 | **8344 / 8344** |
| `aieventtableinfo` | 12 B (composite) | 0 / 988 | **988 / 988** |

All 9,332 carry the ordinary `[id][u32 name_len][name][NUL]` with `name_len == 0`.

### Blast radius, measured not asserted

Every entry header on a live 1.16 install, decoded with the old code and the new:

| | |
|---|---|
| headers compared | 289,905 |
| byte-identical | 280,573 |
| changed | **9,332** — exactly the two tables above |
| id == index key **before** | 280,573 / 289,905 |
| id == index key **after** | **289,905 / 289,905** |

Widths 1 and 2 already used `key_size`; for width 4, `key_size` **is** 4. So the generalisation cannot affect anything that already worked, and the numbers confirm it.

### Tests

`tests/test_entry_header_wide_keys.py` (new, 12 tests) proves it against committed fixtures for both tables, including a test asserting the old u32 read matched **zero** entries — so anyone narrowing it back sees why it was widened.

The old `test_wide_keys_still_use_the_u32_branch` pinned the behaviour this PR corrects. It is replaced, not deleted, and states why the original reasoning was wrong so it cannot quietly come back.


AgentKush:

```
mercenaryinfo        21/21 named   Mercenary_Main, Vehicle_Horse, Vehicle_Dragon
mercenarygroupinfo   11/11 named   Mercenary_Main, Vehicle, Wagon
socketinfo             2/2 named   DaeilBandSocket_Parts_Artifact, Equip_Jewel_Socket
autospawnfilterinfo    1/1 named   DefaultFilter

21,481 live entry headers compared old-vs-new    0 differences
 4,094 fixture entry headers compared            0 differences
```

(`gamestartinfo` still needs #340 to get an index at all, as described.)

Issue #329, opened 2026-07-29
[Mod]: Character Creator still not showing the race pop-up when importing the zip file


lurkser:
### Mod name + Nexus page

Character Creator - Female and Male --- https://www.nexusmods.com/crimsondesert/mods/837?tab=files&file_id=12192

### Mod version

7.7

### Mod type (if known)

Not sure

### Skipped-patch / error text

```text

```

### Your game version

1.15.00

### Does the mod work in another manager (e.g. DMM) on the same game version?

None

### CDUMM bug report

```text

```
As the title said, using CDUMM 3.7.0, but this time even extracting the mod and manually importing it's parts crashed the game, and I see no error message in CDUMM.
Also tested on a fresh install with just this mod and had the same issue.

faisalkindi:
Thanks for testing this on a clean install, @lurkser — that detail rules out a lot on its own and is genuinely useful.

I had a specific theory and I've already disproved it, so recording that here so nobody spends time re-running it. v3.7.0 shipped new characterinfo writes whose byte layout was proven against a game 1.13 table, and you're on 1.15. Another fix in the same batch exists precisely because a different table's record size shifted between game versions, so a stale layout being written into 1.15 looked like a strong candidate for a silent crash.

It isn't that. I extracted characterinfo from a live 1.15 install and ran the shipped parser over it: 7,105 records indexed, the first 600 all parsed cleanly and all passed the safety gate that decides whether CDUMM will write to a record at all. So that path is behaving correctly on 1.15 rather than writing to guessed offsets.

Which means I can't yet explain your crash, and I'd rather say that than guess again.

Two things would let me get at it:

**1. The bug report.** The CDUMM bug report box in your issue came through empty. In CDUMM, Tools then Report a Bug generates one — it lists your game version, which mods are applied, and the last chunk of the log. That's the fastest route to the actual cause, especially since you're seeing no error message.

**2. Which part crashes.** You mentioned extracting the zip and importing the parts by hand also crashed. If you still have that extracted folder, importing the pieces one at a time and noting which one causes it would narrow this a lot. Character Creator ships a model swap, icons and a JSON patch, and those go through completely different code paths — knowing which is at fault changes where I look entirely.

[...]

Issue #347, opened 2026-08-04
CharacterInfo is modelled as 14 fields of 164, so current mods that set character_scale / character_weight apply nothing


AgentKush:
Found by sweeping a **fresh corpus of 55 mods updated in the last 3 days** (many versioned `CD1.16.01/02/03`) through `scripts/apply_scan.py --game-dir` against live build 24545272. These are mods authors have *just rebuilt for the current patch*, so they are the population that should work.

Exit code 6 — six distinct drop classes, 30 dropped intents. **27 of the 30 have one root cause.**

## The root cause

`schemas/pabgb_type_overrides.json` gives `CharacterInfo` an `_ordered_fields` list of **14** entries. That override *replaces* the schema rather than extending it (the loader says so in as many words: ``# _ordered_fields REPLACES the upstream schema's array order.``), so CDUMM loads CharacterInfo as a 14-field type:

```
_isBlocked _stringKey1 _stringKey2 _enum4hash_a _enum4hash_b _description
_spawnFixType _isRemoteCatchable _keyLookup_a _keyLookup_b _vehicleInfo
_callMercenaryCoolTime _callMercenarySpawnDuration _mercenaryCoolTimeType
```

The base schema knows **164** fields for the type. The other 150 — including the two current mods actually write — are unreachable:

```
_characterScale    in base schema: yes    in loaded schema: NO
_characterWeight   in base schema: yes    in loaded schema: NO
```

Most of those 14 are mercenary / vehicle / spawn plumbing, the rest generic (`_isBlocked`, `_stringKey1/2`, `_description`). None of them is appearance, scale or weight, which is what the popular mods edit.

## What it costs, measured

| dropped | mod | intent |
|---|---|---|
| x24 | "Damiane 2x Taller camera" family | `characterinfo` → `character_scale` on `Damian` / `Damian_Clone` |
| x3 | Character Creator (#837 v7.8) | `characterinfo` → `character_weight`, `f36` on `Kliff_Clone` |
| x1 | Mask No Drop (`supports: 1.16`) | `dropsetinfo` → `list[N].raw_16` / `raw_24` |

The engine's own words:

```
Format 3 mod 'camera': no supported intents for characterinfo.pabgb, 2 skipped.
  Mod produced 0 byte changes.
characterinfo: could not locate field 'character_weight' for entry 'Kliff_Clone'
  (record parsed only partially), skipping
[...]

PR #350, opened 2026-08-07
characterinfo: f36 is the gender byte, not the parser's _f36 u32

faisalkindi:
Character Creator / Female Animations (#302) has **never worked once**, on any CDUMM version. woowoots confirmed that this week: the `.field.json` has always stopped the game from starting, while the mod's raw offset version worked perfectly until 1.16.0 broke its hardcoded offsets.

That rules out the #329 all-or-nothing refusal as the cause. The refusal is recent; the crash predates it and survived both behaviours.

Following it back, the `f36` mapping is wrong. It was added for this exact mod and was never confirmed in game.

## Evidence

**1. Field correspondence.** The mod ships the same edit twice: this Format 3 file, and a raw offset patch with hand-written labels. The two agree field for field on all five records (7/6/6/3/3). Every field pairs off by exact value except one. Once the rest are matched, the leftover is `f36` here and `_gender` there, both set to 2.

| Format 3 name | real field |
|---|---|
| appearance_name | _upperActionChartPackageGroupName |
| character_prefab_path | _lowerActionChartPackageGroupName |
| skeleton_name | _characterGamePlayDataName |
| lookup_24 | _skeletonName |
| lookup_25 | _skeletonVariationName |
| **f36** | **_gender** |

**2. Position.** That patch puts `_gender` one byte at block+66. On the live table that byte still reads `01` on Kliff, Kliff_AI and Yann, exactly the before value its `01 -> 02` edit expects.

**3. Table-wide shape.** Across all 7244 records block+66 holds only 0, 1 or 2 (1357 / 4909 / 978), an enum present on every record. `_f36_offset` is published on 142 records, 2% of the table.

**4. Intent.** Read as gender, the mod sets 1 -> 2 on every record it touches. Read as `_f36`, it writes 2 over an existing 2 on Kliff_AI and PlayerAll, making two of four edits no-ops. No author writes that on purpose.

## Effect

Before: gender was never set, so the part of the mod that actually switches the character never happened, and an unrelated u32 got 2 written into it.

After: the gender byte is written where the working patch writes it. Being block-relative also makes it placeable on every record, so records the post-block walk cannot reach are no longer refused over this field. Kliff_Clone is still abandoned, now solely for `character_weight`.

## Tests

Two added: one pins that `f36` lands at block+66 as a single byte, carries the intent value, and is never a same-value no-op (the signature of a wrong slot). One pins the enum shape of block+66 so a future layout change is caught.

Full fast suite: 2844 passed. The 5 failures on this branch reproduce identically on master and are unrelated (1 is the known `py` launcher case, 4 are storeinfo tests reading live game files, see note below).

## Not confirmed in game

I cannot launch the game, so this is not proven fixed. A test build is going to the reporter. **This should not merge until someone boots it.**

## Unrelated, but found while running this

The 4 storeinfo failures are real and not caused by this change: `const byte at record offset 42 is 0 (expected 1) ... record is not the verified disc-0 shape for layout 'CD 1.16' or the layout has drifted again`. The game updated again (exe dated 5 August) and storeinfo drifted past what the parser models. It fails safe by refusing rather than writing wrong bytes, but store mods are likely dead on the current build. CI cannot see this because it has no game install. Filing separately.

### DIFF
--- a/src/cdumm/engine/characterinfo_writer.py
+++ b/src/cdumm/engine/characterinfo_writer.py
@@ -67,7 +67,33 @@ _FIELD_MAP: dict[str, tuple[str, int, str, int]] = {
     # and is refused by name below -- never written at a guessed position.
     "default_action_action_index": (
         "_defaultActionActionIndex_offset", 0, "<I", 4),
-    "f36": ("_f36_offset", 0, "<I", 4),
+    # `f36` is the record's GENDER byte, at block+66. It is NOT the u32 the
+    # parser publishes as `_f36_offset` (GitHub #302).
+    #
+    # That earlier mapping was added for this very mod and was never
+    # confirmed in-game -- the mod has never once worked. Four independent
+    # measurements say it is the wrong slot:
+    #
+    # 1. Field correspondence. Character Creator ships the same edit twice,
+    #    as this Format 3 file and as a raw offset patch with hand-written
+    #    labels. The two agree field for field on all five records
+    #    (7/6/6/3/3). Every field matches by exact value except one; once
+    #    the rest are paired off, the leftover is `f36` here and `_gender`
+    #    there, both set to 2.
+    # 2. Position. That patch puts `_gender` one byte at block+66. On the
+    #    live table that byte still reads 01 on Kliff / Kliff_AI / Yann,
+    #    exactly the "before" value the patch expects for its 01 -> 02 edit.
+    # 3. Table-wide shape. Across all 7244 records block+66 holds only
+    #    0, 1 or 2 (1357 / 4909 / 978) -- a clean enum on every record.
+    #    `_f36_offset` is published on 142 records, i.e. 2% of the table.
+    # 4. Intent. Read as gender the mod sets 1 -> 2 on every record it
+    #    touches. Read as `_f36` it writes 2 over an existing 2 on Kliff_AI
+    #    and PlayerAll, so two of the four edits would be no-ops -- not
+    #    something a mod author writes on purpose.
+    #
+    # Block-relative also means it is placeable on every record, so records
+    # the post-block walk cannot reach are no longer refused over this field.
+    "f36": (_BLK, 66, "<B", 1),
     # `character_weight` is the SAME SLOT under a different DMM name, and
     # that is documented by the mod author rather than inferred. Character
     # Creator 7.5 shipped as raw offset patches with hand-written labels;
--- a/tests/test_characterinfo_appearance_fields.py
+++ b/tests/test_characterinfo_appearance_fields.py
@@ -240,7 +240,11 @@ def test_refusal_reason_is_reported_not_just_logged(table):
     assert len(refusals) == 1, refusals
     assert "Kliff_Clone" in refusals[0]
     # It must say what could not be placed, or it isn't actionable.
-    assert "f36" in refusals[0]
+    # Only character_weight now: `f36` is the gender byte at block+66,
+    # which is placeable on every record, so it is no longer part of why
+    # this one is abandoned.
+    assert "character_weight" in refusals[0]
+    assert "f36" not in refusals[0]
 
 
 def test_a_lone_half_writable_record_is_abandoned_too(table):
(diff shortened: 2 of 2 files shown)


AgentKush:
Independent confirmation from the executable, arrived at without reference to the mod — so this is a genuinely separate line of evidence rather than the same inference restated.

I have been extracting CharacterInfo's layout from the game's own reflection data: each field has a per-field Korean deserialization error string (`CharacterInfo의 _<field>를 읽어들이는데 실패했다`), each string is referenced exactly once by a RIP-relative `lea` inside the deserializer, and the read that precedes it states its own width. Read-only on the exe throughout — opened and mmapped, never copied, patched or executed.

`_gender` is there, and it is a single byte:

```
idx 29  _aiScriptPathFocusHash   primitive  4 bytes   struct +180
idx 30  _playerTargetableType    primitive  1 byte    struct +184
idx 31  _playerLockOnType        primitive  1 byte    struct +185
idx 32  _gender                  primitive  1 byte    struct +186   <--
idx 33  _personalityInfo         sub-reader           struct +188
```

Three points where this lines up with what you found:

**It is one byte.** Not a u32. The deserializer sets the read size inline immediately before the vtable call, so this is the width the game itself reads, not an inference from the data.

**It sits in a run of three consecutive single-byte enums** at struct +184 / +185 / +186. That is exactly the shape your table-wide measurement describes — a small enum present on every record, values 0/1/2 across all 7244. A `_f36` u32 would not sit in that run.

**It is a real reflected field on CharacterInfo**, so it is not a name invented to fit this mod.

Your evidence is mod-correspondence plus live-table shape; mine is the binary's own reflection metadata and disassembly. Two methods that share no assumptions landing on "one gender byte" is about as good as this gets short of booting the game.

Fully agree on not merging until someone boots it. Neither of our methods can prove the game *accepts* the write — only that the field is where we both think it is. Worth being precise that what is corroborated here is the field identity and width, not the end-to-end fix.

One caution on my own data, so it is not over-trusted: I have verified field *membership* and *width* this way, but I have **not** established a reliable read *order* from it. I cross-checked my derived order against the seven tables with round-trip-verified `_ordered_fields` and it disagrees on several — conditionally-read fields have their error strings emitted in cold blocks at the end of the function, so sorting by address puts them in the wrong place. So take the `_gender` width and identity above as solid, and disregard anything that would depend on my ordering until that is fixed.

Nothing here contradicts the PR. Flagging the limitation because it is adjacent enough to be mistaken for support it cannot give.

faisalkindi:
This is the strongest kind of corroboration available here, because the two routes share no assumptions. Mine reads a mod and a data table and infers the field by elimination. Yours reads the game's own reflection strings and the width the deserializer sets before the call. Neither could have produced the other's answer by accident.

One thing worth stating precisely so nobody later reads more agreement into it than exists: your `struct +186` and my `block + 66` are not the same number and should not be expected to match. Yours is an in-memory struct offset where strings are pointers; mine is a file offset into the serialized record where they are inline. What agrees is the identity and the width, one byte of gender, and the shape of the neighbourhood, a run of single-byte enums. That is the part that matters and it is enough.

Thank you also for flagging the ordering limitation unprompted. Nothing in this PR depends on read order, so it does not touch the change, but volunteering the boundary of your own data before anyone asks is what makes the rest of it usable.

Since you posted, one more piece landed from the live table. With the mapping corrected the writes come out as:

```
Kliff.f36        off=  251   01 -> 02
Kliff_AI.f36     off= 8327   01 -> 02
PlayerAll.f36    off=43789   01 -> 02
```

Kliff's write lands on offset 251, which is byte for byte where Character Creator's own raw offset patch writes `_gender`, on a table where the surrounding fields have drifted by 122, 370 and 681 bytes. Gender sits ahead of the variable length stretch so it did not move. All three are now real 01 to 02 edits; the two silent no-ops the old mapping produced are gone.

Still not merging. Everything above establishes that the field is where we both think it is, and nothing establishes that the game accepts the write, which is the only thing that actually closes #302. A test build is ready for the reporter.

Separately, I closed #351. You were right and I confirmed it independently: those tests read an untracked local snapshot from June rather than live game files, and they call the fixed-offset entry point while its layout parameter now defaults to CD 1.16, so a 1.11 era table is parsed as 1.16. CD 1.11 puts the const at 34, CD 1.16 at 42, which is exactly the byte in the error. My "the 5 August build drifted" was a cause I asserted without checking. It is test hygiene, not a regression.

AgentKush:
Two things, both about my own evidence rather than about the change.

## 1. The ordering caveat I attached is now resolved

I closed my earlier comment with a warning not to over-trust my read order, and said to disregard anything depending on it until it was fixed. It is fixed, in #354, and the fix confirms the caveat was warranted rather than over-cautious.

The cause was the one I guessed: conditionally read fields have their error block **outlined** to the end of the function, so sorting the error-string xrefs by address drops them to the end of the table. Ordering by the hot-path branch that *reaches* the block instead fixes it. On ItemInfo — 101 fields shared with a verified order, by far the strongest oracle here — the address sort diverges at index 9 and the corrected one matches all 101 exactly.

Worth being precise about why my earlier "it agrees with CDUMM's verified tail" was not enough to have caught this: that was one table's four-field tail, and the tail is exactly the part outlining does not disturb. It was a real check that happened to be blind to this failure.

**Nothing in this PR ever depended on read order** — the claim is field identity and width — so this does not change the PR either way. I am noting it because I asked for a caveat to be applied and it should now be lifted.

## 2. Re-verified on the current build, which is two newer than my numbers

My figures here were taken on buildid **24568997** (2026-08-05). The install is now **24613230** (exe 2026-08-07). Re-derived from scratch on the new exe, with strict attribution:

```
idx  32  _aiScriptPathFocusHash   primitive  4 bytes   struct +180
idx  33  _playerTargetableType    primitive  1 byte    struct +184
idx  34  _playerLockOnType        primitive  1 byte    struct +185
idx  35  _gender                  primitive  1 byte    struct +186   <--
idx  36  _personalityInfo         sub-reader           struct +188
```

**The struct offsets and widths are byte-identical to what I posted above.** `_gender` is still one byte at struct +186, still the third of three consecutive single-byte enums.

One difference I should point at rather than let someone else notice: the **indices** moved, 29/30/31/32/33 to 32/33/34/35/36. That is not a build change — it is the ordering fix in item 1 pulling three conditionally read fields back to their correct earlier positions. The identity, the width and the neighbourhood are unchanged; only my numbering of them improved. Since the numbering was the part I told you not to trust, that is the expected shape of the correction.

Also unchanged on the new build, for the two fields #347 is about:

```
_characterWeight   struct 0x1E0 (480)    consumes 4 bytes
_characterScale    struct 0x3E8 (1000)   consumes 4 bytes
```

Pattern-A widths validated against the base 164-field schema: **76 agree, 0 disagree.** Slightly fewer comparisons than the 78 I reported before, because the strict attribution leaves 7 of 190 fields unattributed rather than guessing them — I would rather that number go down than have it hold up by including attributions I cannot defend. Still zero conflicts.

## What this does and does not do for the PR

It does not move the thing that actually blocks merging. Both of our methods still only establish that the field is where we think it is; neither shows the game *accepts* the write, and that remains the only thing that closes #302. Your position that this waits for someone to boot it is still right.

What it does do is remove the one asterisk I had put on my own contribution, and confirm the evidence holds across two further game builds — which is a slightly stronger claim than when I posted it, since a values-only hotfix could in principle have moved the struct and did not.

All exe work remains read-only: opened for reading and mmapped, never copied, patched or executed.

Issue #351, opened 2026-08-07
storeinfo tests use a legacy fixed-offset path against a stale local fixture (not a regression)


faisalkindi:
Found while running the full suite against a current install, not from a user report. Filing it before someone hits it.

**The storeinfo parser no longer reads most of the store table on the current game build.**

```
StoreinfoParseError: const byte at record offset 42 is 0 (expected 1) at byte 114;
record is not the verified disc-0 shape for layout 'CD 1.16' or the layout has
drifted again
```

Scope, measured rather than estimated:

```
only 102 entries round-tripped (expected >=260)
```

So roughly a third of the table still parses. The rest is refused.

Four tests catch it, all of them reading live game files:

- `test_storeinfo_native_parser.py::test_live_entry_3101_round_trips_byte_exact`
- `test_storeinfo_native_parser.py::test_live_full_file_clean_entries_round_trip`
- `test_storeinfo_writer.py::test_hernandpets_applies_end_to_end`
- `test_storeinfo_writer.py::test_new_record_with_unmapped_field_refuses`

The game exe on the install I tested is dated 5 August, so this is drift from a build after the one the `CD 1.16` layout was pinned against.

**Two things worth saying plainly.**

It fails safe. The parser refuses rather than walking off and writing to a position it cannot account for, so the failure mode is store mods reporting no changes rather than corrupting the table. That is the behaviour we want, and it is why this is a "mods stop working" problem and not a "saves get damaged" problem.

CI cannot see this. These tests need a real game install, and the runner has none, so master is green while this is broken. That is worth fixing separately: a break that only reproduces on a machine with the game is a break nobody notices until a user reports it, which is exactly the shape of the iteminfo break on 1 August.

Affected mods are the ones editing shop stock and refinement costs. Not yet confirmed against a specific mod on Nexus, since I found this from the test suite rather than from a failing mod.

Related: the same class of problem on iteminfo was #337 / v3.9.0, and characterinfo is currently in the same state under #302 and #350.

AgentKush:
Same build here, and I can narrow where the drift is **not**.

## Build confirmed

The install I verified is buildid **24568997**, exe and every `.paz` rewritten 2026-08-05 10:10 — the 5 August exe you describe. So we are looking at the same thing.

## The entry layer is intact — the drift is below it

Read live, without the parser:

```
storeinfo   entries 432   key_size 2   body 866,221 bytes
[...]

PR #352, opened 2026-08-08
WIP: storeinfo stock-list count offset 44 -> 45 on CD 1.16.03 (#351)

AgentKush:
**Draft — not ready to merge.** The direct parse measurement and the repo's own layout scorer disagree, and I could not reconcile them. Opening so someone who knows `locate_stock_list` can settle it quickly.

Addresses #351.

## What is measured

Driving `parse_stock_list` over all 432 live entries on buildid 24568997 (exe 2026-08-05):

```
count offset   parsed   non-empty   in-entry   round-trip
   44 (old)       80         0          80         80
   45            418       310         418        418
   16            430         0         430        430
```

**Offset 44 parses only entries whose stock count is zero** — 80/80 of its successes are empty. The record walk never runs, the const-byte tripwire at record offset 42 is never reached, and no store containing stock parses at all.

That reframes #351: "102 of ~293 parse" is not a third of the table working, it is a set of vacuous passes on empty stores. Store mods are entirely non-functional on this build, not partially.

Offset 16 is the same trap with a larger number — most entries parsed, every one empty. **Only 45 reads real stock**, and all 418 of its parses round-trip byte-exact through `serialize_stock_list`, which is the strong oracle rather than a plausibility guess.

Record-internal offsets (34/38/41/42) are unchanged: they are measured from the record start, which the count position derives, so a one-byte shift of the count carries them along.

Count distribution at 45 also looks like a store table — median 6 items, p75 12, max 226, 108 empty — versus 44's median of 1794 and max of 4294967295 (`0xFFFFFFFF`).

## Why this is a draft

```
test_detection_beats_every_other_candidate
AssertionError: assert (397, 6376) < (397, 6376)
```

`_score_layout` rates the old 44 layout and the new 45 layout **identically**, while the direct parse above separates them 0 vs 310 non-empty.

I cannot reconcile that with confidence. The likely explanation is that `locate_stock_list` searches for the list rather than trusting `count_payload_offset`, so the scored path and the path I measured are not the same path — in which case this constant may not be the live mechanism and the real fix is elsewhere. That is exactly the kind of thing that is obvious to whoever wrote it and expensive for me to guess at.

I would rather flag the contradiction than talk myself past it. A change that looks convincing on one metric and neutral on another is how a wrong constant gets shipped.

## Also included

Two label assertions loosened from `== "CD 1.16"` to `.startswith("CD 1.16")`, so a point-release layout does not fail a test that meant "the 1.16 family". Those are the same brittle-constant pattern the parser itself keeps hitting.

## Verification

```
storeinfo suite   44 passed, 6 skipped, 1 failed (the scorer tie above)
```

Before this change the suite had 4 storeinfo failures on this build; with it, 1 — but I am explicitly not claiming that as success given the scorer disagreement.

All read-only against the game; nothing written to any install.

### DIFF
--- a/src/cdumm/engine/storeinfo_native_parser.py
+++ b/src/cdumm/engine/storeinfo_native_parser.py
@@ -157,6 +157,28 @@ class StoreLayout:
 #: Newest first -- detection prefers the current game, and an older build
 #: only wins if it actually decodes better.
 LAYOUTS: tuple[StoreLayout, ...] = (
+    # CD 1.16.03 (buildid 24568997, exe 2026-08-05): the stock-list COUNT
+    # moved one byte, 44 -> 45. Nothing inside the record moved -- the
+    # record-relative offsets below (34/38/41/42) still hold, because they
+    # are measured from the record start, which the count position derives.
+    #
+    # Measured on the live table, all 432 entries (GitHub #351):
+    #
+    #     count offset   parsed   non-empty   in-entry   round-trip
+    #        44 (old)       80         0          80         80
+    #        45            418       310         418        418
+    #        16            430         0         430        430
+    #
+    # 44 now yields only EMPTY stock lists: it parses 80 entries and every
+    # one has count 0, so the record walk never runs and the const-byte
+    # tripwire is never reached. That is why the failure looked like "a
+    # third of the table still parses" -- those 80 are vacuous passes, and
+    # no store containing stock parsed at all.
+    #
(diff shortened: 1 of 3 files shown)


AgentKush:
Closing: the +45 change is a no-op. locate_stock_list scans for the list rather than using count_payload_offset, which is why _score_layout rated both layouts identically. On the real path CD 1.16 already locates 397/432 entries and 6,376 records with zero not-found — the whole table. Details in #351.

PR #360, opened 2026-08-11
Derive and prove on-disk layouts for undecoded tables — 14 tables tile exactly, 193 shared readers are the real bottleneck

AgentKush:
Groundwork toward the ~116 tables CDUMM cannot read (Refs #347, roadmap §1). **Nothing is wired into the schema by this PR** — it derives, proves, and reports. The schema change belongs in its own reviewed PR, on top of this evidence.

## The blocker is width, not order

#354 gives field *order*. Screening every undecoded table with it tiles **zero** of them, and the failure is always the same — the walk ends **short**. The shipped schema has no usable width for most of their fields: **3** of them for `actionpointinfo`'s 8.5 MB, **110 of 205** for `gimmickinfo`. One missing width ends a walk no matter how right the order is.

Measured across the 94 undecoded tables that have a reflection class, 1,274 fields:

```
Pattern A (exe states the width)   475  (37%)
Pattern B (typed sub-reader)       636  (50%)
unattributed                       163  (13%)
tables that are Pattern A throughout:  0
```

So nothing unlocks on order plus inline widths. **But the sub-readers are heavily shared** — 193 distinct ones gate all 94 tables, and `sub_1411A31F0` alone is called by **79** of them. That turns 116 reverse-engineering jobs into one problem with a long tail, which is the actual strategic finding here.

## Two stages, nothing accepted unproven

**Stage 1 — static.** Solve each sub-reader recursively. The correction that mattered: a subcall performing **no stream read consumes zero bytes**. Requiring every helper (hash lookups, allocators, loggers) to have a model solved 2 of 193 readers instead of 5. Readers containing a backward branch are refused outright — that's a count-prefixed list and its element width isn't derivable this way.

Read off the disassembly, then independently corroborated by [NattKh's `PABGB_DECODE_PROCESS.md`](https://github.com/NattKh/CRIMSON-DESERT-SAVE-EDITOR-AND-GAME-MODS/blob/main/CrimsonGameMods/PABGB_DECODE_PROCESS.md):

| reader | consumption | evidence | tables |
|---|---|---|---|
| `sub_1411A31F0` | `4 + n` | u32 len, `test edx,edx` early-out at 0, then `mov r8d,[len]` — their doc: "CString reader: u32 len + bytes[len]" | **79** |
| `sub_141071DE0` | `13 + n` | 1B + 8B, then calls `sub_1411A31F0` | 18 |
| `sub_141297490` | `4` | `mov r8d,4` + hash lookup storing a `u16` — their doc: "u32 (file) → u16 (memory)" | 17 |
| `sub_141296000` | `4` | byte-identical shape | 8 |

**Stage 2 — by constraint.** No game file states a list element width. I checked: `.debug$P` is an array of code pointers rather than CodeView type records, and no shipped asset carries a schema (1.69M indexed files, nothing). But the **data** constrains it, so E is searched under `ORDER_ELEM_SIZE`'s own rule — *exactly one size satisfies every record, or the answer is unknown*. Solving one reader can make another table single-unknown, so this iterates to a fixed point.

## Result on buildid 24613230

```
sub-readers solved statically      5 of 193
list widths solved by constraint   7
PROVEN: 14 tables, 1,536 records, exact tiling on 100% of records
```

`aidialogtypeinfo` · `aimemoryinfo` · `breakableobjectinfo` · `categorygroupinfo` · `failmessageinfo` · `gameplayvariableinfo` · `gameversiondatainfo` · `gimmickeventtableinfo` · `materialmatchinfo` · `materialrelationinfo` · `socketgroupinfo` · `terrainregionnaviinfo` · `vibratepatterninfo` · `zoneinfo`

The gate is **exact tiling**, not "reached the last field" — a walk can run out of fields with bytes to spare and has then understood nothing.

## The ambiguous case is the proof this is honest

`dialogvoiceinfo` accepts **all 128** candidate widths, because every one of its lists is empty so nothing constrains E. It is reported as **AMBIGUOUS and left unknown** rather than taking the first hit. That distinction is the whole reason the other 14 are trustworthy.

## What this does NOT establish

**Semantics.** Tiling proves record *structure* and gives every named field its byte offset. It does not prove `_limitDistance` means a distance. Nothing here is writable until values are cross-checked and gated to `_verified_fields` — the community stat mapping was wrong on 18 of 24 entries, all flagged "verified", which is why that gate exists.

**Known limits:** 188 of 193 readers are refused for containing a loop, and the constraint search only handles tables with exactly **one** unknown reader — which is why the cascade reaches a fixed point in two rounds. Extending it to two unknowns (128², fine for small tables) and making stage 1 loop-aware are the next two levers.

## One thing worth knowing separately

I diffed [CrimsonDesertModdingTools](https://github.com/NattKh/CrimsonDesertModdingTools)' schemas against ours: **they are already identical** — 434 tables both, the same 362 reader addresses, zero tables we lack. The roadmap's "347 types" is stale. Also its README says MIT but the repo has **no LICENSE file**, worth knowing before anything from it is ever vendored.

`capstone` + `pefile` are analysis-only, deliberately not app dependencies. Read-only against the install throughout. `ruff` clean.

Refs #347


AgentKush:
I've pushed two more commits since opening this, so here's what changed and why — I'd rather this thread carried the reasoning than leave it buried in the diff.

## b2b0116 — searching two unknown readers at once

I'd capped the constraint search at one unknown reader per table, which meant most tables never got examined at all. Widening it to two took me from 14 tables to 16, and turned up four more readers:

```
sub_141299700 = 4 + count*2     sub_14129A9D0 = 4 + count*2
sub_14129B0E0 = 4 + count*4     sub_1412AD940 = 4 + count*20
```

New tables were `crafttoolgroupinfo` (12 records) and `factionreblockadinginfo` (108).

The part I think is actually worth having is how I decide a reader is settled. My first cut was all-or-nothing on the *pair*, and that throws information away — one reader can be genuinely pinned while the other is unconstrained, which happens when the second one's lists are all empty in that table. So I check each reader separately and pin it only when *every* tiling solution agrees on it. Both got pinned on both new tables, but the mechanism matters for the tables where they won't.

On cost: 128×128 would be slow if I ran every combination against every record, so I prune with a spread of up to 24 records and then re-verify each survivor against the full table. That's sound rather than a shortcut — a width that tiles all records must tile any subset, so a subset failure can't be a false rejection. What I report is always the full-table result.

I also fixed a log line that said "1 of 2 unknowns" even when both had been determined. On a tool like this, a message that overstates what was pinned is the same category of problem as guessing.

## 46e22e7 — sized stream reads, and localstringinfo

Two things I had wrong in stage 1.

I was treating any loop as making a reader undecidable. That's only true when the loop is on a path that reads the stream — allocators and string builders all loop, and I was letting them disqualify their callers.

I was also counting any `call qword ptr [rax + 8]` as a stream read. It isn't. A real read is a **sized** one — a width in `r8d`, *then* the vtable call. That's what kept `sub_141071DE0` refused: the helper it calls, `sub_1410741B0`, makes exactly that call on `r13 + 0x20`, which is a container, not the stream.

Both fixes were right and **neither was enough** — `sub_141071DE0` is still refused by stage 1. So rather than keep bending the analyser around one function, I read it off the disassembly (1 byte, 8 bytes, then the CString reader, so 13 + n) and put it in a new `HAND_VERIFIED` table.

I've kept that table deliberately separate from the automatic models and labelled it as asserted, with the disassembly it came from and the tiling that proves it. An unmarked hand constant is how a schema stops being trustworthy, and you already have `schema_verify` because "someone said so" turned out wrong on 18 of 24 entries. It's held to the same gate as everything else.

That one entry is what unlocked `localstringinfo`:

```
before   16 tables,  1,656 records
after    17 tables, 45,519 records
```

43,863 of those are `localstringinfo` alone — 2.29 MB, the localisation string table, and the sort of thing translation mods actually want.

One caveat I put in the code: those addresses are build-specific. On a different build they shift, and a stale entry then just fails to match any callee rather than applying a wrong width to the wrong reader. That's the failure direction I wanted.

## Where I think this actually stands

17 of 135 tables, 7 of 193 readers. The shape of the progress is the interesting bit — 0 → 14 → 16 → 17 tables, but 1,536 → 45,519 records. One high-fanout reader (`sub_1411A31F0`, used by 79 tables) and one big table did most of the work, which says the next real gains are in the highest-fanout *refused* readers rather than in breadth.

`dialogvoiceinfo` is still reported ambiguous and left alone — all 128 widths tile because every one of its lists is empty. Nothing in that table can settle it; only another table using the same reader could.

AgentKush:
One more commit — `c96cb5f`. This one doubles the statically-derived readers but **doesn't** move the table count, and I want to be upfront about that rather than let the headline number imply more than it did.

## What I changed

The reader convention passes the stream in `rcx`. So a subcall can only consume stream bytes if `rcx` still points at the stream when the call happens. Where `rcx` has been reloaded from memory or a rip-relative global, it's some other object entirely and the callee can't touch our stream no matter what's inside it.

That's exactly how the hash-lookup helper gets invoked:

```
mov  rcx, qword ptr [rip + 0x50646a4]   ; a global map, not the stream
add  rcx, 0x68
call 0x140312c70
```

I'd been counting that as a possible stream read, and it's what kept `sub_141297490` refused. That one reader blocks **10 tables** — the largest single blocker in my ranking, which is why I went after it.

```
readers solved statically:  6  ->  13   (of 193)
```

I put the same tracking in both `reads_stream` and `solve_reader`, since the two have to agree or you get a reader that's judged decidable by one and not the other.

## Why that didn't become more tables

The tables sitting behind those readers have *other* unknowns too. My constraint search only engages a table when at most two readers are unresolved, so anything with three still waits. So this is groundwork — it moves readers out of the unknown set, which is what eventually brings those tables into range, but on its own it unlocks nothing new.

Still 17 tables, still 45,519 records, `dialogvoiceinfo` still correctly ambiguous. Nothing regressed.

## Where I'd go next

Looking at the ranking after this change, the remaining blockers are much flatter — no single reader blocks more than about 3 tables now that `sub_141297490` is done. That changes the shape of the work: there's no big lever left, it's a long tail.

There are also 20 tables that each need exactly **one** more reader, which is the cheapest place to keep chipping:

```
activitypresetinfo  aimovespeedinfo  detectdetailinfo  detectinfo
fieldlevelnametableinfo  gameadviceinfo  materialblooddecalinfo
partprefabdyeslotinfo  relationinfo  sequencerspawninfo  socketinfo  ...
```

Several of those want `sub_141275420` or `sub_1412962B0`, each blocking 2–3 tables. Reading those off the disassembly one at a time is unglamorous but it's the honest route, and each one is a ~20-instruction function.

AgentKush:
`4dfd2cb` — this one's a bug I introduced, and fixing it moved things a lot more than the careful work did.

```
readers solved statically:  13 -> 47   (of 193)
tables proven:              17 -> 22
records proven:         45,519 -> 53,886
```

## What I'd got wrong

Stage 1 disassembled a fixed 600-byte window starting at each reader's entry point. Most of these readers are short — well under that — so my scan just carried on into whatever function happened to sit next in memory, and then blamed the reader for the neighbour's loops and calls.

`sub_1412656D0` is the clean example. It's a 4-byte read followed by a tail-call thunk. Nothing else. It came out as "contains a loop, refused".

I want to be clear that this was worse than lost wins. Bytes from an unrelated function could have contributed a sized read and given me a **wrong width**, and the only thing standing between that and a bad schema entry would have been the tiling gate. I'd rather not generate wrong models and lean on the gate to catch them.

`body()` now stops at the first `ret` that no forward branch jumps past — a `ret` sitting before the furthest branch target is one arm of the function, not its end.

New tables from it: `characterappearanceindexinfo` (8,344 records), `factionoperationgroupinfo`, `gameadvicegroupinfo`, `houseinfo`, `socketinfo`.

## The bit I think is actually worth your attention

Two commits ago I added a `HAND_VERIFIED` table and put `sub_141071DE0 = 13 + n` in it, because stage 1 refused that reader and I'd read it off the disassembly myself. I made a point at the time of labelling it as asserted rather than derived.

With the scan bounded, **stage 1 derives that reader on its own** — along with 34 others. The hand model was never filling a gap in the method. It was covering for this bug.

I've removed the entry. I kept the mechanism, empty, with a note saying the first thing to suspect about a reader stage 1 can't derive is stage 1. That's the actual lesson and I'd rather it be in the code than in my head.

If you'd told me an hour ago that the fastest route to +34 readers was a disassembly bounds bug rather than more RE, I'd have argued. Worth recording.

## Gate still doing its job

Two tables now report AMBIGUOUS instead of a width — `dialogvoiceinfo` and `gamestartinfo`. Both accept all 128 candidates because every one of their lists is empty, so nothing in either table constrains the answer. Neither is claimed.

## Where that leaves it

22 of 135 tables, 47 of 193 readers. Still a long way from everything, but the ratio moved in the right direction and the remaining blockers are flatter than before. I'll keep working the single-unknown tables.

AgentKush:
`2093cbd` — a small one, and it came with a near-miss I want on the record.

```
readers solved statically:  47 -> 48   (of 193)
tables proven:              22         (unchanged)
```

## What I added

Not every backward branch is a count-prefixed list. Some readers have a fixed-length loop, and its consumption is completely determined:

```
loop: mov r8d, 8 ; call [rax+8]     ; read 8
      inc esi
      cmp esi, 4                    ; trip count, an IMMEDIATE
      jb  loop
```

So a read inside one of those counts `trip` times, once outside. I kept it deliberately narrow — single counter, `cmp reg, imm` governing the branch, below/below-equal. Nested or overlapping loops, or any bound that isn't an immediate, still refuse the whole reader. A partial answer here is a wrong width, not a smaller one.

## Honest about the size

One reader, and the table count didn't budge. I expected more than that. The flat result tells me something useful though: most of the remaining backward branches genuinely *are* count-prefixed lists, so they're stage 2's problem and no amount of extra static analysis will reach them.

## The near-miss

Earlier in this session I classified `sub_1412656D0` as a plain 4-byte reader. I got that from a scan that stopped at the first `ret` — which in that function is the **failure path**, sitting before two more reads and the loop. The real answer is 4 + 4 + (4 × 8).

If I'd put "fixed 4" into `HAND_VERIFIED` on the strength of that read, it would have been a wrong constant sitting in a table specifically meant for verified ones. It only got caught because I disassembled the whole function before asserting anything.

That's now the **second** time the same bounds mistake has produced a plausible-looking wrong answer, in two different places — first in `solve_reader`'s 600-byte window, then in my own classifier. Both go through `body()` now.

I mention it because "it looked like a 4-byte reader" is exactly the kind of thing that would have shipped quietly and been very hard to trace back later.

## Where this stands

22 of 135 tables, 48 of 193 readers, and the remaining work is now mostly count-prefixed lists that need the constraint search rather than more disassembly. `HAND_VERIFIED` is still empty.

AgentKush:
`e5091af` — this one demotes four tables I'd already claimed in this PR, so I want to be direct about it rather than bury it under a headline number.

```
tables proven    22 -> 22   (a different, better-constrained set)
records          53,886 -> 54,436
ambiguous         2 -> 6
```

## What I'd got wrong

Stage 2 only ever tried one shape for an unresolved reader: `u32 count` then `count * E`. Which means a reader stage 1 refused could *only* be explained as a list — and the "unique" in my unique-or-nothing rule actually meant **unique among lists**. I'd been describing it as unique full stop.

It now tries three families — fixed `N`, `N + string`, and `4 + count*E` — with unique-or-nothing applying across the whole space.

## Four tables I called proven were not

```
materialmatchinfo     ('list',1)   and ('str',0)     both tile
socketgroupinfo       ('fixed',10) and ('list',2)    both tile
gameversiondatainfo   ('fixed',89) and ('list',17)   both tile
crafttoolgroupinfo    now 2 unknowns, no longer solved
```

All four tiled 100% of records under the old search and I reported them as proven. They weren't. The data genuinely cannot distinguish a 1-byte list from a bare string on `materialmatchinfo`, and I couldn't see that because I never offered the alternative. `gameversiondatainfo` has **one record**, so it constrains almost nothing — I should have been suspicious of that on its own.

They're reported ambiguous now and left alone, which is what they should have been all along.

## Five gained, on firmer ground

`contentsphaseinfo`, `detectinfo`, `dialogvoiceinfo`, `jobinfo`, `mercenaryinfo`.

`dialogvoiceinfo` looks wrong at first glance and I checked it specifically, because it had been my example of correct refusal. On its own it accepts **130** shapes. It resolves because `sub_14129A310` gets pinned by a *different* table where it is unique, and `dialogvoiceinfo` then tiles with that value. So it's cross-table propagation working as intended — the reader is determined elsewhere and only verified here, not guessed from this table.

## Cost

Two unknowns across ~300 shapes each would be ~90k tilings per table. Each slot is pruned alone against a handful of records first, and survivors are always re-verified against the full table, so the pruning can't create a false positive.

## Why I think this is worth having

The headline is flat and the certainty behind it isn't. I'd rather claim 22 tables that survive a wide search than 22 that only survived a narrow one — and I'd rather find this myself now than have you find it when a schema entry lands on the wrong bytes.

It also means my earlier comments on this PR overstated four of the tables. Treating "tiles 100% of records" as proof was right; treating it as proof *of uniqueness* was not, and that distinction only shows up when you widen the space.

AgentKush:
`fa33841` — nothing changed, and that was the point.

Last commit demoted four tables because I'd only given stage 2 one shape to try, so "unique" quietly meant *unique among lists*. I flagged an obvious follow-up question at the time: is three families the complete space, or would a fourth demote more?

Rather than guess at plausible shapes, I checked against `semantic/pabgb_types`, which documents the format's actual grammar. Two real families were missing:

```
CArray<CString>   u32 count + count * (u32 len + len bytes)
COptional<T>      u8 flag + (T if flag != 0)
```

Both added. I require the COptional flag to be 0 or 1, since a real optional flag isn't an arbitrary byte.

**Result: the same 22 tables, the same 54,436 records, the same 6 ambiguous.** None of the 22 has a second explanation in the wider space.

## What the claim is now

Stronger than two commits ago. The 22 are unique across every shape the grammar documents in its non-nested forms:

| grammar | my family |
|---|---|
| primitives, `[T;N]` | fixed |
| `CString` | str |
| `LocalizableString` | str |
| `CArray<fixed>` | list |
| `CArray<CString>` | slist |
| `COptional<fixed>` | opt |

Small aside that pleased me: `LocalizableString` is documented as `u8 category + u64 index + u32 len + bytes` = **13 + n**. That's exactly the model I derived for `sub_141071DE0` off the disassembly, arrived at independently. Two routes agreeing on a reader is about as good as it gets short of the game telling you.

## Still not a completeness proof

I haven't enumerated nested combinations — `COptional<CString>`, `CArray<substruct>` beyond a flat element width, tagged variants. So the honest reading of "22 proven" is **"survives every shape the documented grammar offers, non-nested"**, not "proven correct".

That's a narrower claim than it sounds, and I'd rather state it than let it get quoted back later as the stronger one. If any of this ever becomes a schema entry, that's the sentence that should travel with it.

AgentKush:
`22eb936` — no change to the table count, but it answered the question I'd been assuming the answer to, and I think that's the more useful outcome.

## What I added

A reader ambiguous in one table is often pinned by another that uses it. So stage 2 now accumulates the set of shapes each table permits for a reader and intersects them across tables. Intersecting constraints is sound in a way that picking a favourite isn't, and anything pinned this way is labelled `[by intersecting tables]` so you can see where it came from.

## It changed nothing, and why matters

```
unresolved readers in a <=2-unknown table       41
  ...appearing in MORE THAN ONE such table       3
  ...appearing in exactly one                   38
```

Intersection has almost nothing to intersect. 38 of the 41 remaining readers are used by exactly one table, so no other table can constrain them.

**That contradicts the assumption I'd been working from all session.** `sub_1411A31F0` is called by 79 tables and `sub_141071DE0` by 18, and I took that to mean reader sharing was the structural feature of this format — I said as much in the original PR description. It isn't. It's a feature of the *common* readers. The long tail is bespoke: one reader, one table.

## What that means for getting past 22

There's no remaining leverage to find, and I'd rather say that than keep looking for a clever shortcut. Each of the ~38 tables still in reach needs its own reader derived, and solving one doesn't help the next.

So growth from here is straightforward but unglamorous: read the function, confirm its shape, let the tiling gate check it. Roughly 38 of those, each a short function. No structural trick is going to collapse that into one move — the last four commits were me looking for one and finding, in order: a bounds bug (+34 readers), a fixed-loop case (+1), a wider shape space (+0 tables but 4 demotions), and now this (+0).

I kept the intersection because it's correct and costs nothing, and it'll start paying if stage 1 ever resolves enough of the tail that the survivors end up shared. Not because it's the way forward today.

## Suggested reading of the numbers

22 tables is where the *automatable* portion of this ends. Getting to 60 is about 38 individual reader derivations at maybe 10-15 minutes each with the gate to check them — real work, but bounded and low-risk, because a wrong model simply fails to tile and never enters the set.

AgentKush:
No commit for this one — I tried something, it didn't work, and I've reverted it. Recording it so the next person doesn't spend the same afternoon.

## What I tried

I grouped the remaining refusals by cause instead of picking at them one at a time, and 38 of 41 came back identical: **an unpinnable loop**. That looked like leverage — one systematic cause, not 38 bespoke problems — and it contradicted what I told you last commit about the tail being bespoke.

The read counts suggested the shape too. `sub_141271480` shows `reads=[4,4,4]` *plus* a loop, so I read that as "fixed prefix, then a count-prefixed list" — a family I'd never offered. And the prefix is free: it's stated inline in the disassembly, so only the element width needed searching. I added `prefix + u32 count + count*E` and `prefix + u32 count + count*(4+len)`.

## Why it didn't work, and what that actually tells us

It unlocked nothing. When I checked why, the answer was much more informative than the attempt:

```
aimovespeedinfo      reader sub_1412656D0   290 candidate shapes   0 tile
activitypresetinfo   reader sub_141271480   290 candidate shapes   0 tile
equiptypeinfo        reader sub_1412962B0   290 candidate shapes   0 tile
```

**Zero shapes tile.** Not "several tile so it's ambiguous" — none. Across fixed widths 0–128, strings, lists, lists-of-strings, optionals, and now prefixed lists.

That rules out the framing I'd been using. These tables aren't blocked by a missing shape family or by insufficient constraint. If no single-reader assignment can make the walk land on the record end, then **something else in my model of that table is wrong** — a field I'm treating as a fixed width isn't, or the field order is off for that table, or the reader is a composite of several reads and lists that no one-parameter family can express.

## Correcting my estimate from last commit

I told you getting to ~60 tables was "roughly 38 individual reader derivations at maybe 10-15 minutes each". **That was wrong and I'd retract it.** These aren't short functions waiting to be read — they're cases where the whole-table model breaks, and each needs the table diagnosed rather than one function decoded. I don't have a defensible estimate for that, which is itself the useful thing to know.

## Why I reverted rather than left it in

The prefix machinery was ~60 lines that changed no outcome. On a tool whose whole value is that you can trust what it reports, unused complexity is a liability — it's more surface for a future bug to hide in, and it implies a capability that isn't doing anything. `HEAD` is back to `22eb936`.

The measurement stands on its own and it's in this comment, which is the part worth keeping.

## Where that leaves the PR

Unchanged and still green as far as I know: 22 tables, 54,436 records, 48 readers, 6 explicitly ambiguous, nothing hand-asserted. What's changed is my read on what lies beyond it — the next step is diagnosing individual tables, not harvesting readers, and I'd want to do one end-to-end before quoting a number for the rest.

AgentKush:
`37e3e28` — this is the one that actually makes the work usable. **CDUMM can now decode 13 tables it previously couldn't read at all.**

```
AIDialogTypeInfo 108   BreakableObjectInfo 15   CategoryGroupInfo 92
DetectInfo 37          DialogVoiceInfo 519      FailMessageInfo 11
GameAdviceGroupInfo 8  GamePlayVariableInfo 56  GimmickEventTableInfo 1068
JobInfo 209            LocalStringInfo 43863    MaterialRelationInfo 11
VibratePatternInfo 28
```

46,025 records. `LocalStringInfo` is the one I'd point at — 43,863 records, 2.29 MB, the localisation string table.

## It's data, not code

The descriptors all use the grammar you already have, so nothing in the engine changed:

```
fixed 1/2/4       -> u8 / u16 / u32
fixed 38          -> [u8;38]
CString           -> CString
9 + u32len+bytes  -> LocalizableString
```

That last one pleased me. The deriver produced "9 bytes then a length-prefixed string" purely from the disassembly. Your grammar documents `LocalizableString` as `u8 category + u64 index + u32 len + bytes`. Both are 13 + n, reached independently, and `consume_bytes` agrees on the arithmetic. Three routes to the same reader.

## I verified through your walker, not mine

Every table was re-walked with `_consume_field_bytes` and `_payload_offset` off a fresh schema load, required to consume each record to its exact last byte. That's a different code path from the deriver, and it caught something real: my derivation puts the payload at `key_size + 4 + name_len` with **no** trailing-NUL skip, so each entry needs `_no_null_skip`. Without it every field lands one byte late and the whole record mis-reads.

**9 of the 22 proven tables did not reproduce their proof through the loader, and I've deliberately left them out** — 5 the loader refuses outright, 4 tile 0 records. An entry that can't reproduce its own proof is worse than no entry. They're follow-up, not shipped.

## `_verified_fields` is empty on all 13, deliberately

Tiling proves the record structure and each field's offset. It proves nothing about what a field *means* — nothing here establishes that `_limitDistance` is a distance. The empty gate keeps these out of the Game Data grid as facts and out of the writers until someone cross-checks values against real records.

There's a test pinning that specifically, because it's the assumption most likely to get quietly relaxed later.

## Checks

- full fast suite **2,903 passed, 42 skipped, 0 failures**
- new `tests/test_derived_table_overrides.py` — 41 checks, no game install needed: each table loads, loads in the declared order, every field has a descriptor the walker knows, `_verified_fields` stays empty, `_no_null_skip` is set, and each note records its tool and record count
- the 8 pre-existing override entries are **semantically identical** — 0 modified, 0 removed. The diff is only large because the file got a consistent 2-space indent; I checked the parsed JSON rather than trusting the diff.

One thing to be aware of: `test_schema_verify`'s oracle grew from 7 tables to 20. That's deliberate — a candidate order source now has to reproduce these 13 as well, which strengthens that gate — and the test spells out both sets instead of asserting a count, since the old one said a change there should be deliberate rather than accidental.

AgentKush:
`94c3f6c` — the end-to-end check paid for itself. **The failure was in CDUMM, not in my derivation.**

```
tables wired in   13 -> 17     records   46,025 -> 54,411
```

## The bug

`_payload_offset` computed the entry-id width as:

```python
eid_size = 2 if key_size == 2 else 4
```

So every width other than 2 got read as 4. And `parser._parse_entry_header` was **already fixed for exactly this** — this copy just never got the same treatment. Its docstring names the cost in your own words: mercenaryinfo has 1-byte keys, so reading the id as u32 *"swallowed three bytes of the name, and the table was written off as 'memory-order misaligned' on the strength of it"*.

`_entry_name` carried the identical line, so it's fixed too — there it returned a garbage name that silently fell back to `intent.entry`.

The width is now just `key_size`, matching the parser. Widening is inert for anything that already worked (widths 1 and 2 already used `key_size` there, and for width 4 `key_size` **is** 4), so only mis-read tables change.

## What it recovered

All four are 1-byte-key tables. They had proven layouts that your walker couldn't reproduce, which is why I left them out of the last commit rather than ship them:

```
characterappearanceindexinfo    69/8344  ->  8344/8344
mercenaryinfo                     0/21   ->     21/21
socketinfo                         0/2   ->       2/2
terrainregionnaviinfo             0/19   ->     19/19
```

## Why I'm glad I verified through your walker and not mine

If I'd checked with the deriver's own walker, all four would have looked fine — it computes the payload correctly. The schema entries would have gone in against a payload offset the apply path calculates differently, and the symptom would have been mods on those tables quietly writing to the wrong bytes.

That's the whole argument for making the verification use a different code path from the derivation, and it's the second time this session that the "prove it twice, differently" habit caught something a single check would have missed.

## Scope beyond these four

This wasn't only about my tables. **Any 1-byte-key table reached through the Format 3 apply path was affected.** `parse_pabgh_index` reports `key_size = 1` for terrainregionnaviinfo, socketinfo and mercenaryinfo on the live build, so it was live — just untested, because no wired-in table had a 1-byte key until now.

## Checks

- full fast suite **2,915 passed, 42 skipped, 0 failures**
- derived-override tests now pin 17 tables (72 checks)
- `test_schema_verify` oracle 20 → 24, spelled out rather than counted
- ruff on `format3_apply`: 25 findings before, 25 after — none added

## Still out

5 tables remain — `aimemoryinfo`, `contentsphaseinfo`, `factionoperationgroupinfo`, `houseinfo`, `zoneinfo`. Different cause: they aren't in `pabgb_complete_schema.json` at all, so the loader never reaches their override. Their layouts are proven; they just need a base-schema entry to hang off. That's next.

AgentKush:
`e89c30b` — the last five are in. **All 22 derived tables are now wired in and verified through CDUMM's own walker.**

```
tables wired in   17 -> 22     records   54,411 -> 54,436
```

`AIMemoryInfo` 3, `ContentsPhaseInfo` 3, `FactionOperationGroupInfo` 9, `HouseInfo` 4, `ZoneInfo` 6.

## Why these needed a different fix

They weren't in `pabgb_complete_schema.json` at all. The loader walks the base schema's tables, so an override for a table with no base entry is never reached and `get_schema` just returns `None` — which is why they failed silently instead of loudly.

I added them with field **names only** (`{"f": name}`), deliberately with no invented `stream`/`type`. The base schema is your IDA-derived inventory; the widths belong in the override. Keeping them apart means a future re-import of the upstream schema can't quietly overwrite a derived width, and it stays obvious which source said what.

One of the five was my own error: I had it as `AiMemoryInfo`, the exe calls it `AIMemoryInfo` — and a name that doesn't match is exactly the silent-refusal case above.

Base schema check: 5 tables added, **0 pre-existing tables modified, 0 removed**, verified by comparing parsed JSON against HEAD rather than reading the diff.

## Checks

- full fast suite **2,930 passed, 42 skipped, 0 failures**
- derived-override tests pin all 22 tables
- `test_schema_verify` oracle 24 → 29, spelled out rather than counted

## The 22, in one place

Order from the exe's reflection error strings by hot-path branch (#354). Widths from the deserializer's sized reads, with list element widths solved against table data under unique-or-nothing. Every table required to consume **100% of its records to the exact last byte**, then re-verified through `_consume_field_bytes` off a fresh schema load — a different code path from the derivation, which is what caught the `key_size` bug two commits ago.

`_verified_fields` is empty on all 22, with a test pinning it. Tiling proves structure and offsets, not meaning.

## Honest ceiling

This is where the static route stops. 22 of the ~116 undecoded tables. The remaining ones aren't blocked by ambiguity or a missing shape — for those, **zero** shapes tile, meaning the whole-table model breaks somewhere and each needs individual diagnosis.

Which is why I think Path C from the roadmap — the ASI hook that watches the engine read its own tables — is worth more than another pass at static analysis. I'm looking at that next.

PR #362, opened 2026-08-12
Table coverage: 7 -> 46 tables proven byte-exact, five wrong layouts retracted, ten derivation bugs fixed

AgentKush:
> **Current state** (this PR grew well past its original scope — the sections below are the first wave, and the commit log plus comments carry the rest):
>
> **7 → 46 tables proven byte-exact, 61,111 records**, every one held to exact tiling on 100% of its records through CDUMM's own `_consume_field_bytes`. Five coincidental layouts caught and retracted; ten bugs fixed in the derivation itself, two of which I introduced and the gate caught before they shipped. Full suite 3047 passed / 43 skipped.
>
> 100% coverage is **not** reachable by this route, for three measured reasons: 4 tables have no reflection class in the exe at all (`entitlementinfo`, `keymap`, `levelinfo`, `reviepointinfo`); the largest remaining table (`actionpointinfo`, 28,958 records) is a recursive tree; and the rest are dominated by unsized reads inside larger structs (26 tables / 15,692 records). Nothing unreached is editable — the grid refuses every field when `verified_fields is None`.

---

Continuing the table-derivation work from #360. Six new tables read record-by-record, two shipped entries **retracted as wrong**, and four bugs fixed in the deriver — two of which were silently capping coverage, and one of which had already put a wrong layout into a release.

Measured on game buildid 24613230. Every number here is from a run, not an estimate.

## The retraction first

`HouseInfo._houseRegionDataList` shipped in v3.13 as `CArray<[u8;46]>` and `FailMessageInfo._failMessageInfoList` as `CArray<[u8;33]>`. **Both are wrong.** Neither element has a constant width — each ends in a string read:

```
sub_1412A8A50 element = sub_14129CD10(4B) + sub_141297490(4B) + sub_1411A31F0 CString  = 12 + n
sub_1412AABF0 element = sub_141296000(4B) + sub_141071DE0 LocalizableString            = 17 + n
```

They fitted because every string in the measured build is one length: HouseInfo's 7 element strings are all 34 bytes (12+34 = 46), FailMessageInfo's 18 are all 16 (17+16 = 33). I walked both tables under the fixed *and* the variable model and **both tile 100% of records exactly**, so no amount of data from this build separates them. The disassembly does.

The root cause is not a slip in two entries. Stage 2 recovers a list element width by fitting table data, and `Deriver.candidates()` offers **only fixed-width list elements** — there is no variable-element list shape in that space at all. So unique-or-nothing reported uniqueness inside a space that never contained the competing hypothesis. Uniqueness within an incomplete candidate set is not uniqueness, and 4 and 11 records are few enough for a constant to land by coincidence.

Corrected by defining the elements as substructs, following the existing `StageInfo_Field608Entry` precedent (`u32 + CString`). I audited every `CArray<>` in the shipped overrides against its reader's disassembly; exactly these two claimed a fixed width the element does not have.

**User impact: none.** Both tables have `_verified_fields` empty, so nothing writes to them and the grid shows no values from them. The entries were wrong, not dangerous.

## Four bugs in the deriver

**`find_index` could never match the `.bak` files its own docstring promised.** The glob was `game_index*.sqlite*`, but the rotated copy is `game_index.<timestamp>.bak` — the timestamp *replaces* the extension. Symptom: "No game index found under %LOCALAPPDATA%/cdumm" on a machine with one sitting right there, which is how I hit it. Candidates are now validated by content (opened and checked for `data_tables`) rather than by filename, since filename matching is what failed.

**25 of the install's 135 tables were being dropped silently.** `main()` picked candidates with `s.lower() in lower` and dropped every stem with no identically-named reflection class. 20 were dropped for nothing but a filename convention — the file omits the `Info` the class keeps (`faction.pabgb` ↔ `FactionInfo`, `board.pabgb` ↔ `BoardInfo`). Those tables were never failing the tiling gate; they were never offered to it, and the count just read 94 and looked complete. `class_for_stem()` tries `stem` then `stem + "info"`, **exact matches only** — no fuzzy matching, because difflib cheerfully offers `QuickSlotInfo` for `equipslotinfo` and `RegionInfo` for `reviepointinfo`, and binding a table to the wrong class hands the walker another type's field order. Candidates: **94 → 112.**

**A proven table was still listed as ambiguous.** `dialogvoiceinfo` appeared under PROVEN with 519 records *and* under AMBIGUOUS with "130 shapes tile". Stage 2 runs to a fixed point and a table ambiguous early is often proven later once another table pins the reader that was free in this one; `ambiguous[stem]` was never cleared when that happened.

**163 of 167 stage-1 refusals were a single cause.** `LOOP_UNPINNED`: `fixed_loops` only pins a loop whose trip count is an immediate, and these are `CArray<T>` readers whose count is read from the stream. But a dynamic count does not imply a dynamic element — the loop body's own reads settle it without touching table data. `Deriver.list_element()` now derives it.

Two things were needed, and I found both by the method **failing its oracle**, not by reasoning:

- **Frame tracking.** The same slot is reached through different registers: `sub_1412A8A50` stores its count via `lea rdx, [r11 + 8]` and bounds its loop with `cmp dword ptr [rsp + 0xa0]`. `r11` was `rsp` on entry and `rsp` has since moved 0x98, so `rsp+0xa0 == r11+8` — one slot, two spellings. `rsp` is tracked through the **prologue only** and then frozen, because a linear sweep cannot tell an epilogue from fall-through code, and `sub_14129A5F0`'s early-return `add rsp,0x38 / pop / pop` shifted every later slot by 0x48.
- **`r8` is general-purpose.** `mov r8, qword ptr [rcx + rdx*8]` in `sub_14129A5F0`'s body is a hash-bucket pointer load, not a read width. Treating any non-immediate `r8` write as an unsized read refused all eight oracle readers.

## The oracle

Stage 2 had already pinned eight of these readers from table data — an independent line of evidence. The static derivation had to agree or be thrown away:

| check | result |
|---|---|
| reproduces the data-derived width exactly | 5 (widths 2, 2, 4, 12, 16) |
| correctly reported **variable** | 2 (the retraction above) |
| abstained rather than guess | 1 |
| contradictions | **0** |
| false positives on the 4 readers proved plain fixed | **0** |

`candidates()` is now reader-aware and withholds the list shapes when the element is variable, so those readers come back *unresolved* — which is correct, since the shape they need is not in this walker's grammar.

## Six tables wired in

Verified through CDUMM's own `_consume_field_bytes`, not the deriver's walker:

| table | records | fields |
|---|---|---|
| `factiongroup` | 7 | 6 |
| `uisocialaction` | 2 | 5 |
| `crafttoolinfo` | 19 | 4 |
| `crafttoolgroupinfo` | 12 | 5 |
| `factionreblockadinginfo` | 108 | 5 |
| `globalgameeventgroup` | 12 | 6 |

**Keyed by file stem**, which is the part that would have failed silently. `_load_schemas` iterates the base schema and does `overrides.get(table_name)` with no normalisation, and the app asks `get_schema()` for the stem `identify_table_from_path` returns. An override keyed `FactionGroupInfo` resolves to `factiongroupinfo` — a key nothing ever looks up for `factiongroup.pabgb` — and would have sat in the file doing nothing while a verification script that passed the class name reported EXACT on 100% of records. My first attempt did exactly that; `has_schema(stem)` is now part of the gate.

## One table withdrawn by the rule, not by hand

`royalsupply` proved, and I withdrew it. Its element reader `sub_1412A1370` calls `sub_1412A11E0`, which is **itself** a count-prefixed list (`4 + count*20`) — so the element varies with the inner count, and the unique `('list', 28)` fit only holds while every inner count is 1. That is the HouseInfo coincidence one level deeper, on a 4-record table. `list_element` now returns `variable` when an element's sub-reader is itself a list instead of abstaining, and `royalsupply` drops out of the deriver's own PROVEN set as a consequence.

The remaining small tables are safe for a reason rather than by luck. `factiongroup` (7), `uisocialaction` (2), `crafttoolgroupinfo` (12) and `factionreblockadinginfo` (108) are **fully static** — every width comes from the disassembly and the data refereed nothing, so a low record count cannot flatter them. `crafttoolinfo` (19) and `globalgameeventgroup` (12) each have one searched width, unique among ~290 shapes across records of differing lengths.

## Verification

```
28 derived entries, exact tiling on 100% of records, 0 failed, 54,596 records
deriver:  candidates 94 -> 112,  PROVEN 22 -> 26 tables,  AMBIGUOUS 6 -> 4
          sub-readers 52 solved + 50 count-prefixed lists pinned from the loop body
full test suite: 2997 passed, 43 skipped
```

Still `_verified_fields: []` on every derived table. Tiling proves where a field sits, not what it means — these are readable, not editable, and nothing writes to them.

## Known-unfixed, measured

`conditioninfo` (10,529 records) is the largest table still refusing, and it is now fully classified so the refusal is real rather than a scan artefact. `solve_reader` assumes the stream arrives in `rcx`; `sub_141EF6800` takes it in `rdx`. That assumption is wrong for at least one reader and I have not fixed it here.

### DIFF
--- a/schemas/pabgb_complete_schema.json
+++ b/schemas/pabgb_complete_schema.json
@@ -28443,5 +28443,79 @@
   {
    "f": "_isBlocked"
   }
+ ],
+ "factiongroup": [
+  {
+   "f": "_isBlocked"
+  },
+  {
+   "f": "_factionGroupName"
+  },
+  {
+   "f": "_factionInfoList"
+  },
+  {
+   "f": "_knowledgeInfo"
+  },
+  {
+   "f": "_uiIconPath"
+  },
+  {
+   "f": "_uiDailyQuestImagePath"
+  }
+ ],
+ "uisocialaction": [
+  {
+   "f": "_isBlocked"
+  },
+  {
+   "f": "_name"
+  },
+  {
+   "f": "_description"
+  },
+  {
+   "f": "_quickSlotUiTabIndex"
+  },
+  {
+   "f": "_quickSlotUiIndex"
+  }
+ ],
+ "factionreblockadinginfo": [
+  {
+   "f": "_isBlocked"
+  },
+  {
+   "f": "_reoccupationQuestDataList"
+  },
+  {
+   "f": "_factionNodeList"
+  },
+  {
+   "f": "_delayTime"
+  },
+  {
+   "f": "_protectCombatPower"
+  }
+ ],
+ "globalgameeventgroup": [
+  {
+   "f": "_isBlocked"
+  },
+  {
+   "f": "_defaultOption"
+  },
+  {
+   "f": "_globalGameEventUpdateOption"
+  },
+  {
+   "f": "_globalGameEventInfoList"
+  },
+  {
+   "f": "_fieldInfo"
+  },
+  {
+   "f": "_updateType"
+  }
  ]
 }
--- a/schemas/pabgb_complete_schema.json
+++ b/schemas/pabgb_complete_schema.json
@@ -28517,5 +28517,22 @@
   {
    "f": "_updateType"
   }
+ ],
+ "royalsupply": [
+  {
+   "f": "_isBlocked"
+  },
+  {
+   "f": "_Quest"
+  },
+  {
+   "f": "_Mission"
+  },
+  {
+   "f": "_defaultRandomList"
+  },
+  {
+   "f": "_stageInfo"
+  }
  ]
 }
--- a/schemas/pabgb_complete_schema.json
+++ b/schemas/pabgb_complete_schema.json
@@ -28557,5 +28557,13 @@
   {
    "f": "_pushKnowledgeToGimmick"
   }
+ ],
+ "factionwaypoint": [
+  {
+   "f": "_isBlocked"
+  },
+  {
+   "f": "_wayPointData"
+  }
  ]
 }
--- a/schemas/pabgb_complete_schema.json
+++ b/schemas/pabgb_complete_schema.json
@@ -28534,5 +28534,28 @@
   {
    "f": "_stageInfo"
   }
+ ],
+ "gimmickgateconnection": [
+  {
+   "f": "_isBlocked"
+  },
+  {
+   "f": "_materialItemInfo"
+  },
+  {
+   "f": "_resultItemInfo"
+  },
+  {
+   "f": "_knowledgeInfo"
+  },
+  {
+   "f": "_srcGateInfo"
+  },
+  {
+   "f": "_destGateInfo"
+  },
+  {
+   "f": "_pushKnowledgeToGimmick"
+  }
  ]
 }
--- a/schemas/pabgb_type_overrides.json
+++ b/schemas/pabgb_type_overrides.json
@@ -1323,7 +1323,7 @@
     }
   },
   "FailMessageInfo": {
-    "_meta_note": "Derived 2026-08-11 from game buildid 24613230 by tools/derive_table_layout.py. Field ORDER comes from the exe's reflection error strings ordered by the hot-path branch that reaches each error block (tools/extract_field_order_win.py). Field WIDTHS come from the sized reads in the deserializer, with count-prefixed list element widths solved against the table data under a unique-or-nothing rule. PROOF: the walk consumes every one of this table's 11 records to its exact last byte, and no other shape in the format's documented grammar does. _stringKey/_key are omitted because the entry header consumes them. _verified_fields is EMPTY on purpose: tiling proves the structure and each field's offset, NOT what any field means, so nothing here may be presented as a fact or written to until its values are cross-checked.",
+    "_meta_note": "Derived 2026-08-11 from game buildid 24613230 by tools/derive_table_layout.py. Field ORDER comes from the exe's reflection error strings ordered by the hot-path branch that reaches each error block (tools/extract_field_order_win.py). Field WIDTHS come from the sized reads in the deserializer, with count-prefixed list element widths solved against the table data under a unique-or-nothing rule. PROOF: the walk consumes every one of this table's 11 records to its exact last byte, and no other shape in the format's documented grammar does. _stringKey/_key are omitted because the entry header consumes them. _verified_fields is EMPTY on purpose: tiling proves the structure and each field's offset, NOT what any field means, so nothing here may be presented as a fact or written to until its values are cross-checked. CORRECTION 2026-08-12: _failMessageInfoList shipped as CArray<[u8;33]>, a FIXED element width, and that was WRONG. The element ends in a LocalizableString read, so it has no constant width; the fixed value only fitted because every element string in this build is 16 bytes (17+16==33). Stage 2's candidate space contains no variable-element list shape, so it could not see the competing hypothesis and reported uniqueness inside an incomplete space. Both models tile 100% of records exactly, so the DATA cannot separate them -- the reader's disassembly can, and does. Now CArray<FailMessageInfo_Entry> (u32+LocalizableString). Exact tiling still holds under the corrected model, verified through CDUMM's own _consume_field_bytes.",
     "_no_null_skip": true,
     "_ordered_fields": [
       "_isBlocked",
@@ -1334,7 +1334,7 @@
       "type": "u8"
     },
     "_failMessageInfoList": {
-      "type": "CArray<[u8;33]>"
+      "type": "CArray<FailMessageInfo_Entry>"
     }
   },
   "GameAdviceGroupInfo": {
@@ -1804,7 +1804,7 @@
     }
   },
   "HouseInfo": {
-    "_meta_note": "Derived 2026-08-11 from game buildid 24613230 by tools/derive_table_layout.py. Field ORDER comes from the exe's reflection error strings ordered by the hot-path branch that reaches each error block (tools/extract_field_order_win.py). Field WIDTHS come from the sized reads in the deserializer, with count-prefixed list element widths solved against the table data under a unique-or-nothing rule. PROOF: the walk consumes every one of this table's 4 records to its exact last byte, and no other shape in the format's documented grammar does. _stringKey/_key are omitted because the entry header consumes them. _verified_fields is EMPTY on purpose: tiling proves the structure and each field's offset, NOT what any field means, so nothing here may be presented as a fact or written to until its values are cross-checked.",
+    "_meta_note": "Derived 2026-08-11 from game buildid 24613230 by tools/derive_table_layout.py. Field ORDER comes from the exe's reflection error strings ordered by the hot-path branch that reaches each error block (tools/extract_field_order_win.py). Field WIDTHS come from the sized reads in the deserializer, with count-prefixed list element widths solved against the table data under a unique-or-nothing rule. PROOF: the walk consumes every one of this table's 4 records to its exact last byte, and no other shape in the format's documented grammar does. _stringKey/_key are omitted because the entry header consumes them. _verified_fields is EMPTY on purpose: tiling proves the structure and each field's offset, NOT what any field means, so nothing here may be presented as a fact or written to until its values are cross-checked. CORRECTION 2026-08-12: _houseRegionDataList shipped as CArray<[u8;46]>, a FIXED element width, and that was WRONG. The element ends in a CString read, so it has no constant width; the fixed value only fitted because every element string in this build is 34 bytes (12+34==46). Stage 2's candidate space contains no variable-element list shape, so it could not see the competing hypothesis and reported uniqueness inside an incomplete space. Both models tile 100% of records exactly, so the DATA cannot separate them -- the reader's disassembly can, and does. Now CArray<HouseInfo_RegionDataEntry> (u32+u32+CString). Exact tiling still holds under the corrected model, verified through CDUMM's own _consume_field_bytes.",
     "_no_null_skip": true,
     "_ordered_fields": [
       "_isBlocked",
@@ -1824,7 +1824,7 @@
       "type": "u32"
     },
     "_houseRegionDataList": {
-      "type": "CArray<[u8;46]>"
+      "type": "CArray<HouseInfo_RegionDataEntry>"
     },
     "_floorHeight": {
       "type": "u32"
--- a/schemas/pabgb_type_overrides.json
+++ b/schemas/pabgb_type_overrides.json
@@ -2150,5 +2150,670 @@
     "_stageInfo": {
       "type": "u32"
     }
+  },
+  "AIEventTableInfo": {
+    "_meta_note": "Derived 2026-08-13 from game buildid 24613230 by tools/derive_table_layout.py, reflection class AIEventTableInfo. Field ORDER comes from the exe's reflection error strings ordered by the hot-path branch that reaches each error block. Field WIDTHS come from the sized reads in the deserializer; count-prefixed list ELEMENT widths come from the loop body via Deriver.list_element (validated against the 8 readers stage 2 had independently pinned from table data: 5 reproduced exactly, 2 correctly reported variable-length, 1 abstained, 0 contradictions). PROOF: the walk consumes every one of this table's 988 records to its exact last byte, and no other shape in the format's grammar does. KEYED BY FILE STEM 'aieventtableinfo' because that is what identify_table_from_path returns and what get_schema is asked for; the class name would resolve to a key the app never looks up on this table. _stringKey/_key are omitted because the entry header consumes them. _verified_fields is EMPTY on purpose: tiling proves the structure and each field's offset, NOT what any field means, so nothing here may be presented as fact or written to until its values are cross-checked.",
+    "_no_null_skip": true,
+    "_ordered_fields": [
+      "_isBlocked",
+      "_showName",
+      "_delegateEventHandler",
+      "_reactionLevel",
+      "_allowTypeFlag",
+      "_eventType",
+      "_isSequencerInterruptEvent",
+      "_eventDelayType",
+      "_isTargetMustExist",
+      "_isMustHandled"
+    ],
+    "_verified_fields": [],
+    "_isBlocked": {
+      "type": "u8"
+    },
+    "_showName": {
+      "type": "CString"
+    },
+    "_delegateEventHandler": {
+      "type": "u32"
+    },
+    "_reactionLevel": {
+      "type": "u32"
+    },
+    "_allowTypeFlag": {
+      "type": "u32"
+    },
+    "_eventType": {
+      "type": "u32"
+    },
+    "_isSequencerInterruptEvent": {
+      "type": "u8"
+    },
+    "_eventDelayType": {
+      "type": "u64"
+    },
+    "_isTargetMustExist": {
+      "type": "u8"
+    },
+    "_isMustHandled": {
+      "type": "u8"
+    }
+  },
+  "UIMapTextureInfo": {
+    "_meta_note": "Derived 2026-08-13 from game buildid 24613230 by tools/derive_table_layout.py, reflection class UIMapTextureInfo. Field ORDER comes from the exe's reflection error strings ordered by the hot-path branch that reaches each error block. Field WIDTHS come from the sized reads in the deserializer; count-prefixed list ELEMENT widths come from the loop body via Deriver.list_element (validated against the 8 readers stage 2 had independently pinned from table data: 5 reproduced exactly, 2 correctly reported variable-length, 1 abstained, 0 contradictions). PROOF: the walk consumes every one of this table's 2025 records to its exact last byte, and no other shape in the format's grammar does. KEYED BY FILE STEM 'uimaptextureinfo' because that is what identify_table_from_path returns and what get_schema is asked for; the class name would resolve to a key the app never looks up on this table. _stringKey/_key are omitted because the entry header consumes them. _verified_fields is EMPTY on purpose: tiling proves the structure and each field's offset, NOT what any field means, so nothing here may be presented as fact or written to until its values are cross-checked.",
+    "_no_null_skip": true,
+    "_ordered_fields": [
+      "_isBlocked",
+      "_worldPosition",
+      "_uiTemplateName",
+      "_uiTextureName",
+      "_uiSmallTextureName",
+      "_uiFilterGroupComponentName",
+      "_uiFilterTextureName",
+      "_uiMapLayerType",
+      "_mapIconType",
+      "_knowledgeInfo",
+      "_gameplayTriggerInfo",
+      "_filterGroupName",
+      "_filterGroupParentInfo",
+      "_zIndex",
+      "_isFlexibleSize",
+      "_isFlexibleIcon",
+      "_isSimpleMaterial",
+      "_tooltipText",
+      "_autoRemoveDistance",
+      "_maxScale",
+      "_minScale",
+      "_lerpIconMinSize",
+      "_lerpMinZoom",
+      "_lerpMaxZoom",
+      "_lerpSize",
+      "_changeScaleRatio",
+      "_filterType",
+      "_isShowTooltip",
+      "_isRegionKnowledgeIcon",
+      "_isUIMapQuestType",
+      "_isUIMapDebugQuestType",
+      "_isUIMapDebugQuestAreaType",
+      "_isUIMapNPCType",
+      "_isUIMapMissionType",
+      "_isActorType",
+      "_isPerspectiveIcon",
+      "_isAlwayShowMinimap",
+      "_checkHasOwnerActorIcon",
+      "_isFixScaleIconImage",
+      "_useChangeScale",
+      "_useChangeScaleWhenZoomOut",
+      "_useAutoAbyssLayer",
+      "_minimapForceUpdateIcon",
+      "_indoorStateForceShow",
+      "_otherSpaceForceShow",
+      "_isKeepShowByCharacterInfo",
+      "_isDiscoverGimmickIcon",
+      "_uiFilterGroupByInfo"
+    ],
+    "_verified_fields": [],
+    "_isBlocked": {
+      "type": "u8"
+    },
+    "_worldPosition": {
+      "type": "[u8;12]"
+    },
+    "_uiTemplateName": {
+      "type": "u32"
+    },
+    "_uiTextureName": {
+      "type": "u32"
+    },
+    "_uiSmallTextureName": {
+      "type": "u32"
+    },
+    "_uiFilterGroupComponentName": {
+      "type": "u32"
+    },
+    "_uiFilterTextureName": {
+      "type": "u32"
+    },
+    "_uiMapLayerType": {
+      "type": "u8"
+    },
+    "_mapIconType": {
+      "type": "u32"
+    },
+    "_knowledgeInfo": {
+      "type": "u32"
+    },
+    "_gameplayTriggerInfo": {
+      "type": "u32"
+    },
+    "_filterGroupName": {
+      "type": "LocalizableString"
+    },
+    "_filterGroupParentInfo": {
+      "type": "u32"
+    },
+    "_zIndex": {
+      "type": "u32"
+    },
+    "_isFlexibleSize": {
+      "type": "u8"
+    },
+    "_isFlexibleIcon": {
+      "type": "u8"
+    },
+    "_isSimpleMaterial": {
+      "type": "u8"
+    },
+    "_tooltipText": {
+      "type": "LocalizableString"
+    },
+    "_autoRemoveDistance": {
+      "type": "u32"
+    },
+    "_maxScale": {
+      "type": "u32"
+    },
+    "_minScale": {
+      "type": "u32"
+    },
+    "_lerpIconMinSize": {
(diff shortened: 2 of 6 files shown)


AgentKush:
Pushed a790c87 — `candidates()` is now symmetric, and the reason is a wrong layout the tiling gate let through.

I put this pipeline through an adversarial review after opening the PR, and it found the mirror image of the HouseInfo bug I fixed above. `factionrelationgroup._relationGroupList` gets pinned `('fixed', 24)` **uniquely** on 5 records. But `sub_1412C3ED0` is a count-prefixed list — `mov r8d, 4` reads a u32 count into `[rsp+0x60]`, `mov r8d, 2` reads 2 bytes per element inside the loop, and the loop bound is that same count. The real shape is `('list', 2)`.

The 24 bytes are constant only because the four counts in each record happen to sum to 4 in every one of the five:

```
Graymane        0,2,0,2      HostileCombat   0,3,0,1
FriendlyCombat  0,2,0,2      NPC_Common      0,1,0,3
Monster_Common  0,4,0,0
```

so consumption is 16 + 2*4 = 24 throughout. Byte +12 of that field is two list elements in one record and a u32 list header in another.

What makes it worth a guard rather than a footnote is that none of the checks I'd reach for catch it. Every ±1 perturbation of 24 fails, so the width looks firmly pinned. The five records agree with each other, so more records from that table wouldn't help either. Only the reader's own code separates the two readings.

So `candidates()` now cuts both ways: it already withheld **list** shapes for a reader whose element is variable-length, and it now also withholds **fixed** shapes for a reader that is a known count-prefixed list.

Worth being precise about what was actually broken and when: stage 1b on this branch already prevented this specific case — it pins `sub_1412C3ED0` as `('list', 2)` before any search runs, which is why my audit of all 28 shipped entries for this class comes back clean. The review measured against `3e47db9`, before stage 1b existed. I'm landing the guard anyway because two of the review's own proposed replacements call `candidates()` with no `va` argument and would quietly discard the protection, and I'd rather the invariant live in the function than in the habits of its callers.

Verified: the screen fires on `sub_1412C3ED0` (0 fixed shapes offered) and not on the four readers stage 2 proved are plain fixed (129 offered each); variable-element readers still get 0 list shapes; `candidates()` with no argument is unchanged at 290 shapes; the deriver's proven set is byte-identical to the previous run at 26 tables / 54,581 records with AMBIGUOUS 4; 111 tests pass.

Two things the review turned up that I have **not** fixed here, recorded so they aren't lost:

- `.pdata` is not a function-bounds oracle. It enumerates unwind ranges, and a function with chained unwind info has a primary entry plus continuations flagged `UNW_FLAG_CHAININFO`. 22 of the candidate tables have their field strings spanning more than one entry — QuestInfo covers 36 across 6 ranges. Anything anchoring on the nearest `BeginAddress` rather than the chain root will mis-bound those functions.
- Read attribution is one-to-many. Several field names can share a single read: `charactergroupinfo`'s `_groupGenderList`, `_groupAgeList`, `_stopAnimConstraintDead` and `_groupTierList` all come from one `call` writing one destination. Any scheme that gives each name its own width is asserting three of every four, and that includes the widths already shipped for immediate-width reads. It needs solving before the field list can be trusted to be complete.

AgentKush:
Pushed three more commits — `1c002f8`, `20de0fd`, `4d1637c`. Two are bug fixes the adversarial review turned up, one opens a bucket I had been treating as a dead end. Derived tables go **29 → 33** (54,781 records).

## `1c002f8` — a backward branch is only a loop if it closes a cycle

`solve_reader` treated every backward branch as a loop and refused the reader. But a branch to an earlier address is only a loop if control can *get back* to it. `sub_141280820` has three separate `jmp 0x141280847` converging on one shared tail — none can reach its own branch again, so none is a loop, but the old test counted three and gave up.

`closes_cycle()` walks the function's own instruction list and asks whether the branch is reachable from its target. Measured:

```
sub-readers   52 -> 67 solved statically
PROVEN        26 -> 27 tables, 54,581 -> 54,697 records
AMBIGUOUS      4 -> 3
gained        materialmatchinfo (116 records)   lost: nothing
```

This reproduces a measurement the reviewer made independently by patching a separate copy of the file, which is why I trusted it enough to implement.

`materialmatchinfo` is worth a note beyond the count: it is the first table to escape a tie the data can *never* break. It was ambiguous between `('list', 1)` and `('str', 0)`, which are the identical byte grammar — u32 length then that many bytes. No quantity of table data separates those, so unique-or-nothing would have reported it ambiguous forever. `solve_reader` now reads `('str', 0)` out of the code instead.

## `20de0fd` — `hot()` ranked outlined fields by the wrong basic block

`field_reads` used `bisect_left(blocks, lea) - 1`; `Func.block_start_of` in `extract_field_order_win.py` — the implementation whose ordering was validated against the hand-RE'd tables — uses `bisect_right`. An **outlined** error block starts *at* its own lea, so `bisect_left` lands one block earlier and reads that block's inbound edges, ranking the field by where an unrelated block is branched from. Fall-through fields are unaffected, which is why it hid.

Measured against the order oracle — the 36 tables whose `_ordered_fields` is proven by exact tiling, which is independent evidence from the disassembly:

```
before  32/36   diverging: CharacterInfo(1) ItemInfo(9) StageInfo(6) WantedInfo(0)
after   33/36   diverging: CharacterInfo(1) StageInfo(6) WantedInfo(0)
```

**ItemInfo** now matches, and that one matters beyond the count: ItemInfo is the table `extract_field_order_win.py` cites as proof of its own ordering, so `field_reads` disagreeing with it was a standing contradiction inside one pipeline. Nothing regressed.

Tiling is unchanged at 27 tables, so this buys correctness rather than coverage — and the reason it can be both is worth stating: tiling checks total bytes consumed, so a reordering inside a run of same-width fields is invisible to it. The order oracle is the check that can see it.

## `4d1637c` — variable-length list elements, shipped as substructs

19 tables / 17,458 records were blocked by one descriptor's limitation, not by anything unknown. `CArray<[u8;N]>` cannot describe a list whose element has no constant width, so `list_element` reported "variable" and stage 2 correctly refused to fit a constant. But the format already expresses this — `SUBSTRUCT_DEFS` has hand-written examples of exactly this shape (`StageInfo_Field608Entry = u32 + CString`). The element was describable all along; nothing was reading it out.

`Deriver.element_parts` now returns the element's members in program order instead of discarding them. Same walk, it just keeps the sequence rather than only the sum.

**The oracle for this is the hand work earlier in this PR.** `element_parts` reproduces both substructs I wrote by hand, member for member:

```
sub_1412A8A50  auto [u32, u32, CString]       hand HouseInfo_RegionDataEntry
sub_1412AABF0  auto [u32, LocalizableString]  hand FailMessageInfo_Entry
```

The script generated its own copies of those two; I diffed them against the hand-written pair, the member types were identical, and then I deleted the duplicates. I also removed the five substructs no override references — an unreferenced definition is dead weight that reads like evidence.

Four tables wired in, each verified through CDUMM's own `_consume_field_bytes`:

| table | records | fields |
|---|---|---|
| `relationinfo` | 52 | 9 |
| `partprefabdyetexturepalleteinfo` | 11 | 3 |
| `royalsupply` | 4 | 5 |
| `fieldlevelnametableinfo` | 2 | 3 |

Two more were built and **rolled back** for not tiling — `materialblooddecalinfo` 0/13, `platformachievementinfo` 0/48 — and their substruct definitions went with them.

**`royalsupply` is back, with a correct model this time.** I withdrew it earlier in this PR because stage 2 had fitted `('list', 28)` on 4 records while its element contains a *nested* count-prefixed list, so 28 only held while every inner count was 1. It is now `u32 + CArray<[u8;20]>`, read from the loop body, with the nested list expressed rather than flattened.

All four are **fully static** — not one width came from a data search. That is what makes a 2-record and a 4-record table acceptable here, and it is the whole lesson of this branch: a searched width on a small table is refereed by data that may agree only by coincidence (HouseInfo, FailMessageInfo, royalsupply's first attempt, factionrelationgroup — four cases now), whereas a width read out of the instruction stream is not refereed by the data at all. The data only confirms it. Record count constrains the former and is nearly irrelevant to the latter.

Member names (`u32`, `name`, `list`, `blob`…) are deliberately anonymous. `element_parts` gives structure and says nothing about meaning, so a name implying meaning would be the same overclaim `_verified_fields` exists to prevent. `_verified_fields` stays empty on all four.

## Verification

```
33 derived entries, exact tiling on 100% of records, 0 failed, 54,781 records
order oracle 33/36, no regressions
oracle intact: the 5 readers stage 2 pinned from data still derive to 2,2,4,12,16
full test suite: 3009 passed, 43 skipped
```

## Where the remaining records are, measured

`actionpointinfo` is the largest table in the install at **28,958 records** and is blocked by two things at once. Reflection names only three body fields for a 232-byte payload, so `_actionPoint` is a ~227-byte struct — and `MAX_ELEM = 128` means even that reading is not in `candidates()`. Raising the cap alone would be actively dangerous: a 227-byte blob tiles 28,887 of 28,958 records and fails the rest, and payload sizes 232/236/240/244/248 step by 4, so the real shape is a struct with an internal list. The fix is generalising this commit's parts-walk from loop bodies to whole function bodies — the same machinery, already validated against the two hand-written substructs.

AgentKush:
Pushed `c3389f6` and `993b964`. One is two stage-1 bug fixes, the other is a wrong field order in a **shipped** table that exact tiling structurally could not have caught. Also included: a full coverage measurement, because I would rather give you the real number than a flattering one.

## `c3389f6` — stage 1 counted register use as a read, and loops it never entered

Two defects, both the same family: a test that was true of the *function* when it should have been true of a *position*.

**A loop disqualified a reader even when no read was inside it.** The gate was "a loop anywhere **and** a read anywhere". `sub_141296960` is one 2-byte read followed by a loop that does a lookup and never touches the stream — it consumes 2 bytes, and it was refused. Loop membership is now per instruction.

**`mov r8, <not an immediate>` was treated as an unsized read wherever it appeared.** r8 is general-purpose: in `sub_141296960` it holds a hash-bucket pointer, in `sub_1411A33C0` a function pointer. That is the same mistake I had already fixed inside `list_element` and had left standing here.

**Fixing the second one naively broke a shipped table, and the gate caught it.** `materialmatchinfo` dropped out of the proven set. Its reader `sub_1411A33C0` had been returning `('str', 0)` — the right answer for the wrong reason. It reads a u32 length and then reads the bytes through `call qword ptr [rax+0x20]` and `call r8`, not the usual `call [rax+8]`, so no unsized read is visible at any call site this scan recognises. The old answer came from the function-pointer load being counted as one.

So the two uses of that flag are now separate, because they want different strictness. The loop veto uses the strict positional notion. CString detection keeps the permissive one — safe *there and only there*, because the rule also requires the reader's first read to be exactly 4 bytes, the length prefix. The loop veto had no such guard, which is exactly why it misfired.

**A third, found while reading output:** `solve_reader` produced `('fixed', 11565661)` for `sub_1440A7CCC` — 11 MB for one field, from a `mov r8d, <big imm>` in a function that is not a reader. Inert today only because `reads_stream` says False so it contributed nothing to any sum. `MAX_READER_BYTES` now refuses it; no record in the install is near that.

```
sub-readers   67 -> 86 solved statically of 219
PROVEN        27 tables / 54,697 records -- byte-identical
AMBIGUOUS     3 -- unchanged
```

No table gained. The 19 newly solved readers belong to tables blocked by something else, so this is reader coverage banked, not records, and I am not claiming otherwise.

## `993b964` — WantedInfo's field order was wrong; two errors cancelled

`WantedInfo` shipped with `_ordered_fields` beginning at `_increasePrice` and with the default NUL-skipping header. **That model tiles 0 of 35 records** — it runs `_useTargetPrice` to offset 16, one past the end of a 16-byte record.

There is no trailing NUL. The name is **empty on every record of this table**, so there is nothing to terminate; the byte being skipped at payload offset 6 is `_isBlocked`, which is the first field of very nearly every table in this format. Corrected to `_isBlocked, _increasePrice, _useTargetPrice` with `_no_null_skip`, which accounts for all 16 bytes of all 35 records: header 6, `_isBlocked` at 6, `_increasePrice` at 7..14, `_useTargetPrice` at 15. **0/35 → 35/35.**

**No behaviour change for any mod**, and this is the part worth being precise about rather than dramatic. `_verified_fields` is `['_increasePrice']`, so that is the only field here a mod can edit, and it sits at offset 7 under *both* models — the old one reached it as "7-byte header + first field", the new one as "6-byte header + 1-byte `_isBlocked`". The two errors cancelled exactly on the one field that mattered, which is why the 2026-07-03 verified-only pass read 30/50/100/500/1500 correctly. The corrected walk reads the same five values at the same offset.

**Exact tiling could not have caught this**: 1+8+1 and 8+1+1 both consume 10 bytes. What caught it was the field_reads **order oracle** — the hot-path branch ordering had WantedInfo as one of 3 tables out of 40 disagreeing with its shipped order, and reading the bytes settled which was right. Under the shipped order every price came out 256× too large (25,600 / 12,800 / 384,000 / 128,000 / 7,680) because the value is shifted one byte.

Order oracle 37/40 → **38/40**. The two remaining divergences are known and tracked: CharacterInfo (14 fields modelled of 164, #347) and StageInfo (header-pinned exe order).

## Coverage, measured rather than estimated

I walked every one of the 135 table files in the install through `get_schema(stem)` and `_consume_field_bytes`:

| | tables |
|---|---|
| tile **exactly** on 100% of records | **34** |
| partial (some records tile) | 4 |
| schema exists but tiles **zero** records | 63 |
| no schema, or no usable fields | 34 |

The 63 need splitting before anyone reads them as 63 bugs:

- **15 are in `DECODED` and have dedicated native parsers** (`iteminfo`, `stageinfo`, `characterinfo`, `regioninfo`, `stringinfo`, `dropsetinfo`, …). The generic walker is not how the app reads those, so zero tiling there is not evidence of a fault, and I have deliberately not touched them on this signal.
- **48 are not in `DECODED`**, so the generic schema is their only decoder and it frames no record correctly — 67,373 records. **None of them has an override from this work**; they are pre-existing base-schema entries whose widths do not add up on this build.

Those 48 are **not a live safety issue**, and I checked the code rather than assuming: `_editable_columns` refuses every field when `verified_fields is None`, which is the case for a base-only table, and `_shape_records` masks from the first field with no decodable width onward. They are unusable rather than dangerous.

## On "100% coverage"

It is not reachable by this method, and the reason is measurable: **4 tables have no reflection class in the exe at all** — `entitlementinfo`, `keymap`, `levelinfo`, `reviepointinfo` — so neither field names nor order can be recovered for them statically. Those need a different source entirely (an ASI hook watching the engine read its own tables, the roadmap's Path C).

The honest position after this branch: 34 tables proven byte-exact where there were 7 at the start, every one held to exact tiling on 100% of its records through CDUMM's own walker, four coincidental layouts caught and retracted, and six bugs in the derivation itself fixed. What is left is enumerated with record counts rather than hand-waved.

## Verification

```
34 tables tile exactly on 100% of their records, 0 failed, 54,816 records
   (33 derived overrides + WantedInfo), all through CDUMM's own
   _consume_field_bytes rather than the deriver's walker
deriver: 27 PROVEN / 54,697 records, AMBIGUOUS 3,
         sub-readers 86 solved + 55 lists pinned from the loop body, of 219
order oracle: 38/40
full test suite: 3011 passed, 43 skipped
```

AgentKush:
Pushed `fe7b62a` — the biggest single win on this branch, and it resolves three of the adversarial review's findings at once. **Shipped tables tiling exactly: 34 → 45** (60,636 records).

## The scan was truncating, not just desyncing

`field_reads` swept from `leas[0] - 0x400`, an arbitrary byte. That is not merely imprecise: **`capstone.Cs.disasm` is a generator that stops at the first byte sequence it cannot decode**, so a desynced sweep does not mis-decode — it *ends*. On 15 tables it ended **before the first field lea**, so no field could be classified. Worse and quieter: with the instruction list near-empty the block map was empty too, so `hot()` fell back to `(lea, lea)` and field **order** silently degraded to exactly the naive lea-address ordering that `extract_field_order_win.py` exists to avoid.

## `.pdata` enumerates unwind ranges, not functions

The review was right that this distinction is load-bearing, and it changed the implementation. A function with outlined blocks has a primary `RUNTIME_FUNCTION` plus continuations flagged `UNW_FLAG_CHAININFO` (0x4), each carrying a trailing `RUNTIME_FUNCTION` pointing at its parent. Anchoring on the nearest `BeginAddress` bounds a **fragment**. I measured this before writing the change:

```
RUNTIME_FUNCTION entries              235,690   (50,662 chained, 21.5%)
tables whose leas span >1 fragment         27   (up to 6, for QuestInfo)
...still spanning >1 CHAIN ROOT             0
tables with a lea outside .pdata            0
merged-extent decode reaches every lea  112/112 tables
```

That last line is the property the old sweep failed on every one of the 15. Following the chain to its root and merging extents is what makes it hold.

## The read attribution was also wrong

The read guarding a field was taken from whatever *physically* precedes its lea. For an **outlined** error block that is a different block's tail, usually an epilogue. `AIEventTableInfo._eventDelayType` is the case: the lookback stepped over `ret` at `0x141271922` into `_isMustHandled`'s `mov r8d, 1` and reported width 1, where the sole branch into the block carries `mov r8d, 8`.

Now, when a lea starts a block with inbound edges, the lookback starts from the **earliest branch into it** — the same hot-path source the ordering already uses — and both the call search and the width search refuse to cross `ret`/`jmp`/`int3`/`ud2`. **Refereed by tiling, not asserted**: width 8 tiles `aieventtableinfo` 988/988; width 1 does not tile at all.

## Result

```
MISALIGNED fields   131 of 131 unresolved -> 7  (all deliberate refusals)
sub-readers         86 -> 91 solved, of 219 -> 242
                    (the sweep exposes callees the truncated scan never reached)
PROVEN              27 -> 38 tables, 54,697 -> 60,517 records, 0 lost
AMBIGUOUS           3, unchanged
order oracle        38/40, unchanged, same 2 known divergences
```

Eleven tables wired in, each verified through CDUMM's own `_consume_field_bytes` rather than the deriver's walker:

| table | records | fields |
|---|---|---|
| `uimaptextureinfo` | 2,025 | 48 |
| `aieventtableinfo` | 988 | 10 |
| `knowledgegroupinfo` | 600 | 13 |
| `questgaugeinfo` | 509 | 13 |
| `gameadviceinfo` | 472 | 15 |
| `categoryinfo` | 432 | 4 |
| `gimmickgateinfo` | 396 | 4 |
| `gimmickgateconnection` | 242 | 7 |
| `equiptypeinfo` | 112 | 17 |
| `questgroupinfo` | 42 | 14 |
| `aiactionattributeinfo` | 2 | 2 |

**All eleven are fully static** — not one width came from a data search, so record count refereed nothing. The evidence is stronger than the counts suggest: `knowledgegroupinfo` has **140 distinct record lengths** across 600 records, `aieventtableinfo` 45 across 988, `uimaptextureinfo` 53 across 2,025, and the model predicts every one exactly.

## Three review findings resolved — each re-checked, not assumed

- **`sub_1412A2130`** was fitted `('fixed', 6)` by stage 2 on data where 485 of 509 records have all-zero trailing bytes. It now derives **statically** as a list with a 21-byte element (parts 13 + 8) — exactly the truth the reviewer read off the disassembly. The fabricated width is gone.
- **`questgaugeinfo`'s `_percent` and `_progressDataList`** were attributed to *one* reader, which was the reviewer's stated reason the field list could not be trusted. They are now distinct: `fixed 8` and the list above.
- **`sub_1412C3ED0`** derives `('list', 2)`, the reviewer's value, so `factionrelationgroup` can no longer be pinned as a 24-byte blob. It is correctly **absent** from the proven set rather than shipped wrong.

## Coverage, re-measured

Every one of the 135 table files walked through `get_schema(stem)` + `_consume_field_bytes`:

| | now | before |
|---|---|---|
| tile **exactly** on 100% of records | **45** (60,636 rec) | 34 |
| partial | 4 | 4 |
| schema exists but tiles zero | 53 | 63 |
| no usable schema | 33 | 34 |

Of the 53, **15 are in `DECODED` with dedicated native parsers** — the generic walker is not their code path, so zero tiling there is not evidence of a fault and I have not touched them on that signal. That leaves **38 tables / 61,795 records** where the generic schema is the only decoder and frames no record correctly (was 48 / 67,373). All are pre-existing base-schema entries; none carries an override from this work, and none is editable — the gate refuses every field when `verified_fields is None`.

## Verification

```
45 shipped entries, exact tiling on 100% of records, 0 failed, 60,636 records
deriver: 38 PROVEN / 60,517 records, AMBIGUOUS 3
         sub-readers 91 solved + 60 lists pinned from the loop body, of 242
order oracle: 38/40
full test suite: 3044 passed, 43 skipped
```

Largest still-blocked table is `actionpointinfo` at 28,958 records: reflection names only 3 body fields for a 232-byte payload, so `_actionPoint` is a ~227-byte struct, and payload sizes stepping 232/236/240/244/248 say it contains an internal list. `MAX_ELEM = 128` cannot express it and raising the cap would be actively wrong — a 227-byte blob tiles 28,887 of 28,958 and fails the rest. It needs the parts-walk generalised from loop bodies to whole function bodies, which is the same machinery already validated against the two hand-written substructs.

AgentKush:
Pushed `4bdc990`. This one is worth **+1 table**, and the interesting part is why — the ceiling behind it is now measured rather than guessed, on both of the biggest remaining tables.

## `reader_parts` — the whole-function analogue of `element_parts`

A field whose reader is a struct with no constant width can now be described as a substruct, the same way list elements already were. Plus one new rule: a **4-byte read immediately followed by an unsized read is a length prefix and its bytes** — a CString. `solve_reader` only recognised that when the 4-byte read was the reader's *first*, which is why a struct with fields before its string was refused.

I measured the reach before writing it. 20 blocked tables decompose (17 without the CString-pair rule, so that rule is worth 3). Most of the 20 were already shipped, leaving 3 to build — and **the gate kept 1**:

| table | result |
|---|---|
| `factionwaypoint` | **ok 475/475** — `_wayPointData` = u32 + u32 + CArray<[u8;12]> |
| `materialblooddecalinfo` | FAIL 0/13 — rolled back |
| `platformachievementinfo` | FAIL 0/48 — rolled back |

Their substruct definitions went with them. **Decomposable is not correct**; two thirds of what this change produced failed exact tiling and was discarded. This line of attack is close to exhausted and I'd rather say so than run another pass over it.

## The ceiling on actionpointinfo — 28,958 records, and it is not a missing feature

Its `_actionPoint` reader `sub_141271150` and `sub_1412B1220` **call each other**. The field is a tree, and a tree is not expressible as a flat substruct at any depth, so `reader_parts` refuses it by design rather than flattening it into something that might tile by luck.

Everything *else* in that reader is now accounted for — fixed 56, 4, list(4), 4, 4, 4, 16, 4, 16, 2, 4, 1 — so the recursion is the entire remainder. Reaching it needs recursive substructs in the walker's grammar, which is a format-model extension, not a scan fix.

## The conditioninfo trap, recorded so it is not walked into twice

10,529 records, 4 fields, **zero unknown readers**, and still no tiling. I hand-read record 0 as `_isBlocked(1) + 8 + CString + 1 = 28` and it fit *exactly*:

```
00 | 03 d7 00 00 00 00 00 01 | 0e 00 00 00 "Weather(Rainy)" | 00
```

So I tested `_gameCondition` at every width from 0 to 16 against all 10,529 records:

```
 8 bytes ->     26/10529        <- what record 0 suggested
 9 bytes ->    746/10529
12 bytes ->   4203/10529        <- best, and still only 40%
```

Nothing tiles. The field is variable-length and my single-record reading was a coincidence. That is precisely the failure this pipeline exists to prevent, it caught me, and the next person will hand-read one record and see the same 8 — so the sweep is written into the task notes as the answer to it.

## Verification

```
46 shipped entries, exact tiling on 100% of records, 0 failed, 61,111 records
oracle intact: the 5 list widths stage 2 pinned from data still derive to
   2, 2, 4, 12, 16; the 3 variable verdicts hold; element_parts still
   reproduces both hand-written substructs member for member
reader_parts correctly REFUSES both halves of the actionpointinfo recursion
full test suite: 3047 passed, 43 skipped
```

## Where the branch ends up

**7 → 46 tables proven byte-exact** (61,111 records), every one held to exact tiling on 100% of its records through CDUMM's own `_consume_field_bytes` rather than the deriver's own walker.

Five coincidental layouts caught and retracted (HouseInfo, FailMessageInfo, royalsupply's first attempt, factionrelationgroup, and WantedInfo's order). Ten bugs fixed in the derivation itself — two of which I introduced and the gate caught before they shipped.

**100% is not reachable by this route, and I can now name all three reasons instead of one:**

1. **4 tables have no reflection class in the exe at all** — `entitlementinfo`, `keymap`, `levelinfo`, `reviepointinfo`. Neither names nor order are recoverable statically.
2. **The largest remaining table is a recursive tree** (above).
3. **The rest are dominated by unsized reads inside larger structs** — 26 tables / 15,692 records, where the width comes from a slot this scan cannot follow.

None of the unreached tables is editable: `_editable_columns` refuses every field when `verified_fields is None`, which is the state of every base-only table, and `_shape_records` masks from the first field with no decodable width onward. They are unusable rather than dangerous, and tasks are open with per-table record counts for whoever picks this up.

The honest one-line summary: the static route has been taken about as far as its grammar allows, and what is left needs either a richer type model (recursive substructs) or a different evidence source entirely — an ASI hook watching the engine read its own tables, which is the roadmap's own Path C and the only thing that can reach the four tables with no reflection data at all.

Issue #363, opened 2026-08-13
Help wanted: macOS field-name/order dump (two text files, script provided — not the game binary)


AgentKush:
**What I need:** two small text files from anyone who owns Crimson Desert **on macOS**. A script is provided; it takes a minute to run.

**What I do NOT need, and please don't send:** the game binary itself, or any game files. That would be redistributing a commercial game, which I'm not asking anyone to do. The script reads *your own* installed copy and writes out a list of field names — a few tens of KB of text.

---

## Why

I've been decoding Crimson Desert's `.pabgb` data tables so CDUMM can read and edit more of them. Current state on the Windows build, all verified by exact tiling on 100% of each table's records through CDUMM's own walker:

- **7 → 46 tables** proven byte-exact (61,111 records)
- five wrong layouts found and retracted, ten bugs fixed in the derivation itself

The method recovers each table's field order from the game's own Korean error strings — `<Class>의 _<field>를 읽어들이는데 실패했다` ("Failed to read `_<field>` of `<Class>`") — ordered by the code that references them.

**The wall I've hit is that the Windows build doesn't name every field.** Measured, not guessed:

| table | records | error strings in the exe | body fields those cover |
|---|---|---|---|
| `actionpointinfo` | 28,958 | 5 (2 are header fields) | 3, for a **232-byte** record |
| `conditioninfo` | 10,529 | 6 | 4, and the record is variable-length |
| `itemuseinfo` | 9,756 | 4 | 2, for a 29–65 byte record |

`ItemInfo` has 115 strings for 113 known fields, so it's fine — but a table with 3 named fields covering 232 bytes cannot be completed from strings that don't exist. About **38 tables / ~62,000 records** are blocked behind some version of this.

## What the macOS build gives that Windows doesn't

Two separate things, and I want both:

1. **Read order.** The same Korean strings exist on Windows (I count **516 classes / 4,421 field references**), but they're pointer-referenced from descriptor structs, so the string table alone isn't ordered — on macOS they're reported to sit in the exact order fields are read. That's `fields.json` below.
2. **The names Windows never emits.** The macOS binary is reported to be **unstripped**. Its symbol table may name the fields that have no error string on either platform — which is the actual blocker for `actionpointinfo` and friends. That's `symbols.txt` below.

If you'd rather not run anything, even just confirming whether the macOS binary is genuinely unstripped (`nm -a <binary> | head`) is useful.

## How to help

[...]

PR #364, opened 2026-08-13
Table coverage wave 3: +2 tables, a field order the hot path got wrong, and a duplicate substruct key

AgentKush:
Follow-on from #362. **48 shipped entries, exact tiling on 100% of records, 61,184 records** (was 46 / 60,636). Full suite 3053 passed, 43 skipped.

## `levelactionpointinfo` (60 records) came free

Re-running the substruct wiring after #362's `.pdata` sweep — which resolved 124 more fields — made its variable list element describable.

## `materialblooddecalinfo`: the hot-path order was wrong

This table had failed the gate twice, and the cause was **not** a width. Every width was already static. The **order** was wrong.

The hot-path branch ranking gives `_isBlocked, _bloodDecalVariationData, _skillKey`, because the list's guard branch sits at `0x14128893F` and `_skillKey`'s at `0x141288960`. That order **tiles 0 of 13 records**, and it cannot be right — reading the list first takes the 4 bytes at payload offset 1 as its element count, i.e. `0x00015F90` = **89,999 elements**. With the list last, record 0 accounts exactly:

```
00                     _isBlocked                1
90 5f 01 00            _skillKey                 4
04 00 00 00            count = 4
  "k6p10" + 16 bytes   element 20 + n   ->  4 x 25 = 100
                                       1 + 4 + 4 + 100 = 109 = rec0
```

and so do all 13.

**So the hot-path ordering is not infallible.** Worth stating plainly, because this line of work leans on it heavily: it is right on 38 of the 40 tables the order oracle can check, and this is a case it gets wrong. Exact tiling is what caught it.

The residual ambiguity is handled rather than ignored. `_isBlocked` (1 byte) and `_skillKey` (4) both precede a variable-length field, so swapping them preserves the total and **both orders tile 13/13** — the same width-preserving blind spot as WantedInfo in #362. The data cannot separate them. Two independent things do: the hot-path ranking, which is reliable for those two (`_isBlocked` at `0x14128891F` precedes `_skillKey` at `0x141288960` — it was only wrong about the *list*), and the format-wide convention that `_isBlocked` is the first field of very nearly every table. Both are recorded in the entry's `_meta_note`.

## Order cannot be recovered by search — tested, and recorded

Before correcting that table by hand I checked whether order could be derived mechanically. For every table whose widths are all static but which does not tile, I permuted the field order under unique-or-nothing, refusing tables with more than 8 fields and refusing any table with fewer than 3 distinct record lengths (a fixed-length table cannot tell permutations apart — that is exactly the WantedInfo trap).

**11 tables tested, 0 recovered.** The blocked tables are blocked by structure, not by order. Recorded so nobody spends the afternoon on it again.

## A real defect ruff caught

`SUBSTRUCT_DEFS` had **`RoyalSupplyInfo_QuestEntry` defined twice** (F601). The two definitions were byte-identical so nothing behaved differently — but a duplicate key silently overrides, and `royalsupply` is a shipped table. Removed, along with four derived substructs no override references (`RelationInfo_GimmickTagDataEntry`, `FieldLevelNameTableInfo_LevelNameInfoDataEntry`, `HouseInfo_HouseRegionDataEntry`, `FailMessageInfo_FailMessageInfoEntry`) — each a stale alias of one that is referenced. An unreferenced definition is dead weight that reads like evidence.

`pabgb_types.py` is back to the 4 ruff findings that are pre-existing on master (import order, SIM103, 2× UP045).

## Verification

```
48 entries, exact tiling on 100% of records, 0 failed, 61,184 records
pytest 170 passed (targeted), full suite 3053 passed / 43 skipped
```

## Where this route stands

The Windows static route is now yielding 1–2 tables per pass, down from 11 in #362. What remains needs a different evidence source, and the routes are ranked with per-table record counts in the tracked tasks. The macOS ask is #363 — it is the only route that supplies the field *names* Windows never emits, which is the actual blocker on `actionpointinfo` (3 named fields for a 232-byte record, 28,958 records).

### DIFF
--- a/schemas/pabgb_complete_schema.json
+++ b/schemas/pabgb_complete_schema.json
@@ -28565,5 +28565,40 @@
   {
    "f": "_wayPointData"
   }
+ ],
+ "uitalktreeinfo": [
+  {
+   "f": "_isBlocked"
+  },
+  {
+   "f": "_rootTalkTreeInfo"
+  },
+  {
+   "f": "_parentTalkTreeInfo"
+  },
+  {
+   "f": "_childTalkTreeInfoList"
+  },
+  {
+   "f": "_conditionInfo"
+  },
+  {
+   "f": "_targetConditionInfo"
+  },
+  {
+   "f": "_buttonText"
+  },
+  {
+   "f": "_npcActivityInfo"
+  },
+  {
+   "f": "_interactionInfo"
+  },
+  {
+   "f": "_effectInfo"
+  },
+  {
+   "f": "_resultDataList"
+  }
  ]
 }
--- a/schemas/pabgb_type_overrides.json
+++ b/schemas/pabgb_type_overrides.json
@@ -2830,5 +2830,43 @@
     "_wayPointData": {
       "type": "FactionWayPointInfo_wayPointData"
     }
+  },
+  "LevelActionPointInfo": {
+    "_meta_note": "Derived 2026-08-12 from game buildid 24613230 by tools/derive_table_layout.py, reflection class LevelActionPointInfo. Field ORDER from the exe's reflection error strings ordered by the hot-path branch that reaches each error block. Field WIDTHS from the sized reads in the deserializer; count-prefixed list element widths from the loop body via Deriver.list_element. PROOF: the walk consumes every one of this table's 60 records to its exact last byte. KEYED BY FILE STEM 'levelactionpointinfo' because that is what identify_table_from_path returns and what get_schema is asked for. _stringKey/_key are omitted because the entry header consumes them. _verified_fields is EMPTY on purpose: tiling proves the structure and each field's offset, NOT what any field means. _levelActionPointGroupList is a list whose ELEMENT has no constant width, so it is CArray<LevelActionPointInfo_LevelActionPointGroupEntry> rather than CArray<[u8;N]>. The element's members come from the loop body's own read sequence via Deriver.element_parts, which independently reproduces the two substructs written by hand for HouseInfo and FailMessageInfo. Member names are anonymous on purpose: the parts give structure and say nothing about meaning.",
+    "_no_null_skip": true,
+    "_ordered_fields": [
+      "_isBlocked",
+      "_accessType",
+      "_levelActionPointGroupList"
+    ],
+    "_verified_fields": [],
+    "_isBlocked": {
+      "type": "u8"
+    },
+    "_accessType": {
+      "type": "u8"
+    },
+    "_levelActionPointGroupList": {
+      "type": "CArray<LevelActionPointInfo_LevelActionPointGroupEntry>"
+    }
+  },
+  "MaterialBloodDecalInfo": {
+    "_meta_note": "Derived 2026-08-13 from game buildid 24613230 by tools/derive_table_layout.py. Widths are static: _isBlocked u8 and _skillKey u32 from sized reads, and _bloodDecalVariationData from Deriver.list_element, which gives a variable element of 20 + n (a CString plus 16 fixed bytes). ORDER WAS CORRECTED AGAINST THE BYTES. The hot-path branch ranking put _bloodDecalVariationData BEFORE _skillKey, because the list's guard branch sits at 0x14128893F and _skillKey's at 0x141288960, and that order tiles 0 of 13 records. It cannot be right: reading the list first takes the 4 bytes at payload offset 1 as its element count, i.e. 0x00015F90 = 89,999 elements. With the list last, record 0 accounts exactly -- 1 + 4 + 4 + 4*25 = 109 -- and so do all 13. The remaining ambiguity is that _isBlocked and _skillKey are 1 and 4 bytes before a variable field, so swapping them preserves the total and BOTH orders tile 13/13; the data cannot separate them. Two independent things settle it: the hot-path ranking, which is reliable for these two (_isBlocked 0x14128891F precedes _skillKey 0x141288960 and was only wrong about the list), and the format-wide convention that _isBlocked is the first field of very nearly every table. _verified_fields is EMPTY: this proves structure, not meaning.",
+    "_no_null_skip": true,
+    "_ordered_fields": [
+      "_isBlocked",
+      "_skillKey",
+      "_bloodDecalVariationData"
+    ],
+    "_verified_fields": [],
+    "_isBlocked": {
+      "type": "u8"
+    },
+    "_skillKey": {
+      "type": "u32"
+    },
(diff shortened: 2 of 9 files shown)


AgentKush:
Pushed `fbea309`. **49 shipped entries, exact tiling on 100% of records, 61,359 records** (was 48 / 61,184).

## Every remaining unclassified field had one cause

After #362's `.pdata` sweep, I re-measured why any field is still `?`. Across all 112 candidates: **86 NO_WIDTH, 5 NO_CALL, 0 misaligned** — and all 86 are `mov r8d, <register>`, concentrated in callee-saved registers:

```
r8d, r15d   41      r8d, r14d    2
r8d, r12d   30      r8d, ebp     1
r8d, r13d   12
```

MSVC hoists a literal into a callee-saved register when a deserializer performs many same-size reads, so the constant is defined once and reused. `field_reads` only accepted `mov r8d, <imm>` and dropped every one of them.

`_const_reg` resolves it by **unique definition**: every write to that register anywhere in the (now `.pdata`-bounded) function is collected, and a value is returned only when there is exactly one and it is `mov <reg>, <imm>`. Two definitions mean the value depends on a path this scan does not follow, and the answer is then unknown — the same unique-or-nothing rule the widths and list elements are held to. The result is also range-checked.

Deliberately **not** a backward walk to the nearest assignment: the nearest in address order is not necessarily the one that reaches, and taking it would be exactly the plausible guess this pipeline exists to refuse.

```
unclassified fields      86 -> 44
tables fully classified  84 -> 98 of 112
resolved widths          1 x433, 4 x175, 8 x26, 2 x10, 12 x8
PROVEN                   38 -> 39 tables, 60,517 -> 60,692 records
order oracle             51/54, unchanged        lost: nothing
```

The width being overwhelmingly **1** matches the two earlier independent readings of this same pattern, which is the cross-check I wanted before believing it.

`uitalktreeinfo` crosses the line — 175 records, 11 fields, verified 175/175 through CDUMM's own `_consume_field_bytes`, **fully static** (no width from a data search) across 12 distinct record lengths.

One table for 42 resolved fields is a poor ratio and I won't dress it up: the other 41 fields belong to tables blocked by something else.

## A change I made, measured, and reverted

The same hoisted-constant pattern plausibly affects the **sub-reader** walkers too — `solve_reader`, `list_element` and `reader_parts` all accepted immediates only. So I extended `_const_reg` to all three. The oracle stayed fully intact (5 list widths, 3 variable verdicts, 4 reader models) and nothing regressed.

And the full pipeline came out **byte-identical**: 91 readers solved, 60 lists pinned, 39 tables, 60,692 records. **Zero change.**

So the 55 `UNSIZED_READ` sub-readers are *not* blocked by hoisted constants — their width genuinely comes from a memory slot (length-prefixed data), which no amount of constant propagation recovers. I reverted it rather than shipping it, because code with no measured effect is complexity without benefit. Recording it here so the next person doesn't spend the same afternoon proving it again.

## The blocker profile now

Of the 91 unresolved field readers:

| cause | count |
|---|---|
| `UNSIZED_READ` — width from a slot | 55 |
| `VARIABLE_ELEM` — element has no constant width | 32 |
| other | 3 |
| decomposable | 1 |

Tables: **53 fully classified / 45 blocked on an unresolved reader / 14 still carrying a `?`**.

That is the honest shape of what is left: the remaining work is reader coverage and variable structure, not field classification — and the `UNSIZED_READ` majority is precisely the class that needs runtime observation rather than static analysis.

AgentKush:
Pushed `16c7c3d`. This one isn't more tables — it's the Game Data grid finally admitting when it can't read one.

## The problem

38 tables whose only decoder is the generic schema frame **no record correctly** on this build — 61,795 records' worth, measured by walking all 135 tables in the install. Nothing there is editable (`_editable_columns` refuses every field when `verified_fields is None`), so it was never *dangerous*. But it still rendered a grid, and a grid built from a layout that is wrong the moment anyone checks reads as data.

## What it does now

`layout_fit()` samples records and reports how far the schema the grid **actually uses** gets through them. Sampled rather than exhaustive so it stays cheap on the display path — `stageinfo` has 51,595 records and the answer doesn't get more true after the two-hundredth — and samples are spread across the table rather than taken from the front, because early records are the least representative (a layout that's wrong about a variable-length field often survives the short ones).

Five states, each saying only what is actually known:

| condition | message | real cases |
|---|---|---|
| walk completes on **no** record | "CDUMM cannot read this table's layout on this game build — the walk stops after about N of M fields on every record sampled. Columns past that point are meaningless" | `stageinfo` 25 of 81, `regioninfo` 21 of 24 |
| completes on **some** | "only reads on N% of records sampled — some rows' columns will be misaligned" | `iteminfo` 86% |
| completes on all, no tiling | "reads consistently, but the schema doesn't account for every byte of a record" | `conditioninfo` |
| completes **and** tiles | "accounts for every byte of every record sampled — the strongest signal CDUMM has that this layout fits your build" | `materialmatchinfo`, `uitalktreeinfo` |
| no schema | unchanged: experimental | — |

The existing `health` note is kept and now sits *below* the structural verdict, because `health` is a heuristic about values looking unusable while this is a measurement of whether the record can be framed at all.

## Two mistakes I made building it

Both caught before shipping, and both are the whole subtlety of the feature — worth recording rather than quietly fixing.

**It first measured `get_schema()`.** The grid uses `schema_for_table()`, which picks the field-order variant that fits the build. Measuring the declared order instead reports **`iteminfo` as unreadable** on any build where a variant is in play — a false alarm on one of the best-supported tables in the app.

**It first used byte-exact tiling as the warning condition.** Tiling is the right gate for deriving and shipping a layout, and the derivation pipeline is held to it. It is the *wrong* test for "should the user trust this grid": a schema that models the first N fields correctly and leaves unmodelled tail bytes shows N correct columns and still fails it. Measured — `iteminfo`, `characterinfo`, `stageinfo`, `regioninfo` and `storeinfo` **all tile 0 of 200 sampled records**. Warning on that would have put a scary banner on the flagship tables.

So the display signal is **completion** (did the walk get through the whole field list), and tiling is kept as a separate stronger property that earns the *positive* message rather than triggering a warning.

## Shape

`layout_fit` lives in `schema_verify.py`, which already imported everything it needs, so it is testable without Qt. It returns `None` — "unknown", never "broken" — when there is no schema or no index, and the call site is wrapped so a failure can never block the view.

```
8 new tests, CI-safe (committed vanilla113 fixture, no game install needed)
full suite: 3064 passed, 43 skipped
ruff: clean on the new file; game_data_page.py and schema_verify.py both at
      their pre-existing counts (36 and 3)
```

Issue #376, opened 2026-08-26
[Bug]: after2.0v cdumm make issue


parkilsooon-lgtm:
### Type of problem

Game crashes (launch or in-game)

### Game install source

None

### CDUMM bug report

```text
None
```

### What did you do, and what went wrong?

environment : mac m5 tahoe

after update crimson desert 2.0 ,
if i run mod,
game make crush and directly turn off

### Environment checks

- [ ] CDUMM is **not** running as administrator
- [ ] The game is **not** installed under `Program Files`
- [ ] The game exe does **not** have the "Run as administrator" compatibility flag set

### Confirmations

- [x] I'm on the latest CDUMM release
- [x] I searched existing issues and didn't find a duplicate

faisalkindi:
Thanks for the report, and sorry about the crash. What you hit is real and understood, and it is not something you did wrong.

The 26 August game update (2.0) changed the layout of several of the game's data files. Two things follow from that:

**Why the game crashes with mods on.** Mod data that CDUMM prepared before the update was built against the old file layout. After the update that data no longer matches the game, and loading it can take the game down exactly the way you describe. We also found and fixed a bug where disabling all your mods could leave that stale mod data on disk anyway (#375, merged today), which would keep the game crashing even after you turned everything off.

**To get the game running right now:**

1. In CDUMM, disable all mods.
2. In your game folder, look for a numbered folder (0036 or higher) that contains a file called `_cdumm_overlay.marker`, and delete that folder. That is CDUMM's mod overlay from before the update.
3. In Steam: right-click the game → Properties → Installed Files → Verify integrity of game files.

The game should start clean after that.

**On re-applying mods:** hold off for now. The 2.0 update changed the item table's layout and CDUMM cannot read it yet, so item mods would apply nothing on this build (#377 tracks it). The fixes from today plus the new layout will ship in the next release, and this issue will get a note when it is out.

One more thing that would help us: you mentioned Mac (M5, Tahoe). If the crash happened with a specific mod list, naming the mods would tell us whether anything beyond the stale-overlay case was involved.

faisalkindi:
v3.16.0 is out and contains everything discussed above: https://github.com/faisalkindi/CrimsonDesert-UltimateModsManager/releases/tag/v3.16.0

[...]

PR #391, opened 2026-08-28
fix(gui): RescanPage never learns when its own rescan finishes

glebkin:
Fixes the Rescan page staying on "Initiating rescan..." forever, with the button never re-enabling, even after the snapshot genuinely completed (the success toast fires correctly, but the page itself never learns).

The page only emits a signal into the main window, which owns the actual snapshot worker, and had no path back to report completion, so the button also never got disabled up front, which is how a user who thought it was frozen ended up triggering a second real rescan by clicking again.

It now disables itself immediately and polls the main window's worker flag, the same technique already used elsewhere for this exact process, to learn when the work is done and refresh its own status and stat cards.

### DIFF
--- a/src/cdumm/gui/pages/tool_page.py
+++ b/src/cdumm/gui/pages/tool_page.py
@@ -11,7 +11,7 @@ import math
 import os
 from pathlib import Path
 
-from PySide6.QtCore import QEasingCurve, QThread, Qt, Signal, Slot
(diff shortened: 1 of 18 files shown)


faisalkindi:
Checked the plumbing: _run_qprocess sets main_window._active_worker for the snapshot QProcess and clears it in the finish handler, so polling it is reliable, and it matches what RecoveryFlow already does. This is the rescan-stuck-forever report from #376 too. Merging, thanks Gleb.

